PR #7, opened 2019-02-14
Static site example

clareliguori:
Example for HTTPS static site infrastructure: S3 bucket, CloudFront distribution, Route53 alias record

### DIFF
--- a/typescript/static-site/README.md
+++ b/typescript/static-site/README.md
@@ -1,0 +1,33 @@
+# Static site
+
+This example creates the infrastructure for a static site, which uses an S3 bucket for storing the content.
+
+The site redirects from HTTP to HTTPS, using a CloudFront distribution, Route53 alias record, and ACM certificate.
+
+## Prep
+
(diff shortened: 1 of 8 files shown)


typescript/static-site/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+## Deploy Site Content
+
+During the infrastructure deployment, you will see an output named something like "StaticSiteBucket34E5D9AF".  The value is the bucket name where you can upload the static site content.
+
+You will also see an output named something like "StaticSiteDistributionId8C64EF2A".  This value is the distribution ID for the CloudFront distribution, which needs to be invalidated each time new content is uploaded to the bucket.
+
+```
+$ aws s3 sync . s3://$STATIC_SITE_BUCKET
```
eladb: You should be able to use the [s3-deployment](https://awslabs.github.io/aws-cdk/refs/_aws-cdk_aws-s3-deployment.html) module to deploy the files instead of `aws s3 sync`. This way, your static site can be packaged as a library and reused.

PR #9, opened 2019-02-15
Doug aws tweak readme

Doug-AWS:
Used example for running TypeScript example.

Added 'cd typescript/my-widget-service' to list of commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ $ npm install -g aws-cdk
 $ cd typescript/my-widget-service
 $ npm install
 $ npm run build
-$ cdk synth   // Optional, to see what CloudFormation will deploy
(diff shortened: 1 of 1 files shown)


rix0rrr:
I just had this discussion with Hsing-Hui.

I don't want `cdk synth` in the list of commands, we'll just end up training users to run a useless extra command every time they want to deploy.

Also, the `cd` in there is rather borderline I feel. Do we really expect people not to understand that?

Doug-AWS:
RE: cdk synth, that's why I added "// optional".

I guarantee you that some user will try running the commands from the root directory. Of course the knowledgeable will think "Well, duh", but so what? It's pedantically correct.

PR #44, opened 2019-05-26
Add an example for AppSync with API Key, and CRUD Resolvers with DynamoDB

simalexan:
*Issue #, if available:*
There wasn't an example demonstrating how to create an AppSync GraphQL API pointing to multiple Resolvers which are doing CRUD operations over a DynamoDB table.

*Description of changes:*
Add a AWS CDK example for an AppSync GraphQL API with an API Key, and CRUD Resolvers with DynamoDB. There are different Resolvers doing two Queries to Get One, All and two Mutations doing Save Item and Delete Item.

Should be useful for developers starting with AWS CDK on how to run a GraphQL API with Resolvers and DynamoDB. .


simalexan:
Along the way, seem as in the example `classic-load-balancer` is failing 
` const vpc = new ec2.VpcNetwork(this, 'VPC');`.
So I fixed it, because of the change here
https://github.com/awslabs/aws-cdk/commit/68efa0443649d897dace22843531f0c373b2fe0e

simalexan:
And another, in the example `lambda-cron` is failing
`events.EventRule` is changed into `events.Rule`. so I fixed that one as well

simalexan:
Added all the fixes for the switch from `ec2.VpcNetwork` to `ec2.Vpc`

simalexan:
Fixed also the ECS `PlacementConstraint.distinctInstances()` and the switch from `loadBalancer.dnsName` -> `loadBalancer.loadBalancerDnsName` (Obviously these weren't normalized when these changes occurred).

simalexan:
Fixed also the step functions example
required a change from `resource: checkJobActivity` to `task: new tasks.InvokeActivity(checkJobActivity)`

That also required to import `@aws-cdk/aws-stepfunctions-tasks`, as per the example in the `aws-cdk`;

rix0rrr:
In the mean time, Elad came by and fixed the examples build. Can you merge again from master?

simalexan:
@rix0rrr thank you for your time, will do that now

Issue #49, opened 2019-06-12
Typescript examples fail to run support scripts with latest CDK releases


NGL321:
**Describe the bug**
The support scripts will no longer build with the latest releases of the CDK. They fail with the below error.

**To Reproduce**
Attempt to run either "scripts/build.sh" OR "scripts/build-npm.sh"

**Expected behavior**
 
```
> cdk@0.1.0 build /codebuild/output/src827049164/src/typescript/elasticbeanstalk/elasticbeanstalk-environment 
> tsc 
 
index.ts(13,27): error TS2551: Property 'getContext' does not exist on type 'ConstructNode'. Did you mean 'setContext'? 
npm ERR! code ELIFECYCLE 
npm ERR! errno 2 
npm ERR! cdk@0.1.0 build: `tsc` 
npm ERR! Exit status 2 
npm ERR!  
npm ERR! Failed at the cdk@0.1.0 build script. 
npm ERR! This is probably not a problem with npm. There is likely additional logging output above. 
 
npm ERR! A complete log of this run can be found in: 
npm ERR!     /root/.npm/_logs/2019-06-11T12_38_31_507Z-debug.log 
 
[Container] 2019/06/11 12:38:31 Command did not exit successfully chmod +x scripts/*.sh && scripts/build.sh exit status 1 
[Container] 2019/06/11 12:38:31 Phase complete: BUILD State: FAILED 
[Container] 2019/06/11 12:38:31 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: chmod +x scripts/*.sh && scripts/build.sh. Reason: exit status 1 
[Container] 2019/06/11 12:38:31 Entering phase POST_BUILD 
[Container] 2019/06/11 12:38:31 Phase complete: POST_BUILD State: SUCCEEDED 
[Container] 2019/06/11 12:38:31 Phase context status code:  Message:  
```

**Version:**
 - OS: OSX-10.13.6 (High Sierra)
 - Programming Language: Typescript
 - CDK Version: 0.34.0

NGL321:
Fixed with PR #50.
Problem may exist for other language bindings. Will investigate and create new issues appropriately

hopewise:
I have the same issue as here:


```
aws-rails-provisioner build
.
.

running npm run build ...
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! cdk-sample@0.1.0 build: `tsc`
npm ERR! Exit status 2
[...]

PR #50, opened 2019-06-13
Update for cdk0.34.0

NGL321:
*Issue #, if available:* #49

*Description of changes:*
Updated typescript examples to account for breaking changes
- Replaced "app.run()" with "app.synth()"
- Replaced "node.getContext()" with "node.tryGetContext()"
- Replaced "route53.AliasRecord()" with "route53.ARecord()"

### DIFF
--- a/typescript/ecs/fargate-service-with-local-image/local-image/app.py
+++ b/typescript/ecs/fargate-service-with-local-image/local-image/app.py
@@ -11,5 +11,5 @@ def hello():
     return html.format(name=os.getenv("NAME", "world"), hostname=socket.gethostname())
 
 if __name__ == "__main__":
-    app.run(host='0.0.0.0', port=80)
+    app.synth()(host='0.0.0.0', port=80)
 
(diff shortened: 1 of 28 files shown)


typescript/static-site/static-site.ts
```diff
@@ -4,6 +4,7 @@ import route53 = require('@aws-cdk/aws-route53');
 import s3 = require('@aws-cdk/aws-s3');
 import ssm = require('@aws-cdk/aws-ssm');
 import cdk = require('@aws-cdk/cdk');
+import targets = require('@aws-cdk/aws-route53-targets/lib');
```
skinny85: This should be just `import targets = require('@aws-cdk/aws-route53-targets');`, similar to the imports above.

typescript/ecs/fargate-service-with-local-image/local-image/app.py
```diff
@@ -11,5 +11,5 @@ def hello():
     return html.format(name=os.getenv("NAME", "world"), hostname=socket.gethostname())
 
 if __name__ == "__main__":
-    app.run(host='0.0.0.0', port=80)
+    app.synth()(host='0.0.0.0', port=80)
```
fulghum: Are the empty parens correct before the args?

NGL321: Actually looks like it was an accidental python replace for Fargate local-image. 
Reverting to "run()".

PR #53, opened 2019-06-19
Create an example of api gateway, lamdda & rds

maskerade:
*Description of changes:*
Example for an API Gateway with a Lambda that queries an RDS Instance. 
Additionally uses a custom resource to populate the DB with test data & lambdas use secrets manager to retrieve the db credentials

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ $ cdk destroy
 
 | Example | Description |
 |---------|-------------|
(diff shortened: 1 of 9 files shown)


maskerade:
Closing this as was tested against v0.34.0 however v0.35.0 has just been released so will test and make any required changes to ensure is compatible with v0.35.0

PR #57, opened 2019-06-24
feat: TypeScript example for Fargate using Route 53 and ACM for HTTPS

ranrotx:
*Issue #, if available:* NA

*Description of changes:*
This example uses Route 53 along with ACM to provision an SSL certificate that can be used by the ALB to enable an HTTPS Fargate service.


NGL321:
I will look into this shortly.
@pkandasamy91 feedback from the ECS side?

PR #64, opened 2019-07-02
Updating Python examples to 0.36.1

### LINKED ISSUE #61: Deprecated aws-cdk/cdk in use
The Python examples (I haven't looked at other languages) use the just-deprecated aws-cdk/cdk module. This has been replaced with the aws-cdk/core module (see [here](https://github.com/awslabs/aws-cdk/blob/master/CHANGELOG.md).)
The fix may well be as simple as replacing all references to aws-cdk/cdk to aws-cdk/core.

garnaat:
*Fixes #61 *

*Description of changes:*

Updating to latest CDK version.  All Python examples have been ported but some of the ECS examples are broken due to https://github.com/aws/jsii/issues/537 which will hopefully be fixed in the next day or so.

### DIFF
--- a/python/ecs/ecs-service-with-advanced-alb-config/app.py
+++ b/python/ecs/ecs-service-with-advanced-alb-config/app.py
@@ -2,11 +2,11 @@ from aws_cdk import (
     aws_ec2 as ec2,
     aws_ecs as ecs,
     aws_elasticloadbalancingv2 as elbv2,
-    cdk,
+    core,
 )
 
-app = cdk.App()
-stack = cdk.Stack(app, "aws-ec2-integ-ecs")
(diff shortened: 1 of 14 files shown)


python/ecs/ecs-service-with-advanced-alb-config/app.py
```diff
@@ -2,11 +2,11 @@
     aws_ec2 as ec2,
     aws_ecs as ecs,
     aws_elasticloadbalancingv2 as elbv2,
-    cdk,
+    core,
```
eladb: We have been importing core types directly lately, so instead of core.Type it’s just Type. Not critical, but that’s what we are doing in TypeScript at least.

PR #76, opened 2019-07-11
CDK General Availability (1.0.0) updates

skinny85:
Fix examples for CDK's General Availability (1.0.0) release.

----------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ node_js:
 python:
     - "3.6"
 
-install: echo Nothing to install
+install: apt-get install python3.6-venv
 script: scripts/build.sh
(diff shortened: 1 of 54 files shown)


skinny85:
We're getting closer...the Java build is succeeding, now only Python is left ;)

rix0rrr:
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.


```

rix0rrr:
Aha, python3.6 is only available starting from Ubuntu 18.04 (`bionic`), newest distro for Travis seems to be `xenial` (16.04).

rix0rrr:
Yay it workses!

skinny85:
Thanks for picking this up Rico!

PR #77, opened 2019-07-13
feat: Reorganize python lambda section and add new example

wulfmann:
*Description of changes:*
Moves existing lambda exalmple to new lambda folder
Adds new lambda with s3 example
Adds new lambda with kinesis stream example


wulfmann:
If you are not looking to move examples into a lambda folder I can remove that part of the PR. I have a few other lambda examples I'm working on which is the only reason I did that.

rhboyd:
I’m not a fan of service-specific directories. How would a new user know this example would be in the lambda folder and not in an S3 folder? I think the example name should make the use-case and/or services used clear

wulfmann:
I do agree with your point there. My thinking was to follow the same convention as the S3 folder. I’ll update these examples for now and if the lambda examples become unwieldy we can revisit.

Thanks

wulfmann:
I've pulled them out of a top-level lambda folder and left them independent.

wulfmann:
Bump. Anything need to be updated on this to get it merged?

python/lambda-with-existing-s3-code/app.py
```diff
@@ -0,0 +1,32 @@
...
+            self, 'LambdaCodeBucket',
+            bucket_name='my-lambda-code-bucket'
+        )
+
+        lambdaFn = lambda_.Function(
+            self, 'Singleton',
+            handler='index.main',
+            code=lambda_.S3Code(
+                bucket=lambda_code_bucket,
+                key='my-lambdaa.py'
```
garnaat: Not a deal breaker but there is a typo there.

wulfmann:
@garnaat thanks! updated

PR #83, opened 2019-07-20
Updating all TypeScript examples for CDK release 1.1.0

curry9999:
*Issue #, if available:*

*Description of changes:*


rix0rrr:
Thanks for contributing, but these fields don't actually correspond to CDK versions. They would indicate the version of the example. Since we're not releasing them to NPM, no need to track  their versions anyway.

PR #84, opened 2019-07-20
Fixes an invalid function name in the lambda-cron Java example

NutterzUK:
***Description of changes:***

**1. Updated unit tests to make sure the assertions were running.**
As one of them didn't actually assert anything (`testTypes`). Whilst there, I also fixed some of the tests and the `actual` and `expected` were the wrong way around, unless using hamcrest's `assertThat` (which it now is).

**2. Fixed deployment by removing the function name.**

**Before**
The function name was invalid, although it passed CFT validation, it resulted in:
```
 3/6 | 11:14:26 PM | CREATE_FAILED        | AWS::Lambda::Function   | SingletonLambdac4f9439a61bd48c9859d3e91a657e2cb (SingletonLambdac4f9439a61bd48c9859d3e91a657e2cbF7A8D345) 1 validation er
ror detected: Value 'CDK Lambda Cron Example' at 'functionName' failed to satisfy constraint: Member must satisfy regular expression pattern: (arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{2}((-gov)|(
-iso(b?)))?-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_\.]+)(:(\$LATEST|[a-zA-Z0-9-_]+))? (Service: AWSLambdaInternal; Status Code: 400; Error Code: ValidationException; Request ID: 67
e1126c-2a1e-4586-a7e6-3e810328682b)
```

**After:**
```
3/6 | 12:05:15 AM | CREATE_COMPLETE      | AWS::Lambda::Function   | cdk-lambda-cron (cdklambdacronE8ADB9B0)
```

### DIFF
--- a/java/lambda-cron/src/main/java/software/amazon/awscdk/examples/LambdaCronStack.java
+++ b/java/lambda-cron/src/main/java/software/amazon/awscdk/examples/LambdaCronStack.java
@@ -7,12 +7,12 @@ import software.amazon.awscdk.services.events.Rule;
 import software.amazon.awscdk.services.events.RuleProps;
 import software.amazon.awscdk.services.events.Schedule;
 import software.amazon.awscdk.services.events.targets.LambdaFunction;
-
 import software.amazon.awscdk.services.lambda.Code;
-import software.amazon.awscdk.services.lambda.Function;
-import software.amazon.awscdk.services.lambda.FunctionProps;
 import software.amazon.awscdk.services.lambda.Runtime;
+import software.amazon.awscdk.services.lambda.SingletonFunction;
+import software.amazon.awscdk.services.lambda.SingletonFunctionProps;
 
+import java.util.UUID;
 
 /**
  * Lambda Cron CDK example for Java!
@@ -21,8 +21,8 @@ class LambdaCronStack extends Stack {
     public LambdaCronStack(final Construct parent, final String name) {
         super(parent, name);
 
-        Function lambdaFunction = new Function(this, "cdk-lambda-cron",
-                FunctionProps.builder()
+        SingletonFunction lambdaFunction = new SingletonFunction(this, "cdk-lambda-cron",
+                SingletonFunctionProps.builder()
                         .withDescription("Lambda which prints \"I'm running\"")
                         .withCode(Code.inline(
                                 "def main(event, context):\n" +
@@ -30,6 +30,7 @@ class LambdaCronStack extends Stack {
                         .withHandler("index.main")
                         .withTimeout(Duration.seconds(300))
                         .withRuntime(Runtime.PYTHON_2_7)
+                        .withUuid(UUID.randomUUID().toString())
                         .build()
         );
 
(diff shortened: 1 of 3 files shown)


java/lambda-cron/src/main/java/software/amazon/awscdk/examples/LambdaCronStack.java
```diff
@@ -7,12 +7,9 @@
 import software.amazon.awscdk.services.events.RuleProps;
 import software.amazon.awscdk.services.events.Schedule;
 import software.amazon.awscdk.services.events.targets.LambdaFunction;
-import software.amazon.awscdk.services.lambda.Code;
+import software.amazon.awscdk.services.lambda.*;
```
rhboyd: I would avoid .* imports for example code. It makes it harder for new developers to know where Classes were imported.

java/lambda-cron/src/test/java/software/amazon/awscdk/examples/LambdaCronStackTest.java
```diff
@@ -8,30 +8,31 @@
 
 import java.io.IOException;
 import java.util.List;
+import java.util.stream.Collectors;
 
-import static org.hamcrest.collection.IsIterableContainingInAnyOrder.containsInAnyOrder;
-import static org.junit.Assert.assertEquals;
+import static org.hamcrest.Matchers.*;
```
rhboyd: same .* import

NutterzUK:
No worries @rhboyd , good idea, I've updated these now.

NutterzUK:
@eladb, Thanks. I've learnt something new, function name is not required  👍 .

**What I'll do:** I'll add another commit removing function name.

**Some things of interest for others from perusing the AWS docs:**
Providing a function name shouldn't cause updating the stack to fail deployment, as long as it is a valid function name, however, [the AWS docs say](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-functionname): "If you specify a name, you cannot perform updates that require replacement of this resource. **You can perform updates that require no or some interruption.** If you must replace the resource, specify a new name."

But then, there are no fields which require replacement of a lambda function other than FunctionName itself, so it doesn't look like it stops anything. 

I believe that the only real problem a function name causes, assuming it is valid, is when deploying mulitple stacks containing the same function name. But if you don't have a specific reason to have a function name, it's best not to provide one which avoids that problem entirely.

One thing I **have** found functionName useful fo,r is that I get an ARN that doesn't change, allowing me to reference it from an alexa skill. I've not yet quite figured out without it how to get the ARN out of the CDK deployed stack and into the `ask cli` without copying and pasting it across, so having it stay the same even if I destroy/re-create the stack is handy, e.g:
arn:aws:lambda:eu-west-1:<account-id>:function:AlexaHandlerLambda
I would imagine we can export this somehow though, that's something for another day for me :).

NutterzUK:
@eladb @rhboyd , thanks for your feedback :). I've updated the pull request and description with the fixes.

Issue #99, opened 2019-08-07
These stacks aren't going to clean themselves


rhboyd:
Example apps should include necessary parameters so that resources are properly cleaned up with with `cdk destroy`. For example, setting S3 RemovalPolicy.DESTROY on all example buckets.

This would be a good first issue for a contributor.

hellnation:
The suggestion is that the standard for aws-sdk-examples should be that roundtripping an example through cdk deploy/destroy shouldn't leave anything in the account wherever possible, and if it does it should be mentioned in the example code comments.

You mentioned DynamoDB as being similarly affected. I will see if there's something similar that can be done for them.

hellnation:
In testing, I note that the behaviour for DynamoDB is different - it will delete the Table even if there is data in it (whereas S3 will error if the Bucket is not empty). I have adjusted the comments next to the code accordingly.

rhboyd:
#100 fixes this. closing.

PR #100, opened 2019-08-08
sets removal policy on S3 and DynamoDB to DESTROY

hellnation:
*Issue #, if available:*
https://github.com/aws-samples/aws-cdk-examples/issues/99

*Description of changes:*
Changes the behaviour of examples that create S3 Buckets and DynamoDB Tables such that these resources are deleted by cdk destroy. Explains the behaviour in comments


eladb:
We should use https://github.com/aws/aws-cdk/issues/2973 once it's available.

PR #155, opened 2019-11-12
fix: Allow ingress traffic from public internet for NLB Ec2/Fargate Service

iamhopaul123:
Fixes [#4794](https://github.com/aws/aws-cdk/issues/4794)

---


typescript/ecs/fargate-load-balanced-service/index.ts
```diff
@@ -20,6 +20,8 @@ class BonjourFargate extends cdk.Stack {
       },
     });
 
+    fargateService.service.connections.allowFromAnyIpv4(ec2.Port.tcp(80));
```
pkandasamy91: We do not want to do this by default. This opens up security concerns.

iamhopaul123: Well, I considered about the security concerns so I opened port 80 instead of opening all ingress ports. Also, I think the example should work as an easy way showing how to make use of the NLB Fargate service (allowing ingress traffic is one of the approaches). The problem of this example is when the stack is deployed and you curl the url it fails to return any value, indicating that the example does not work properly.

pkandasamy91: I agree the example should work, but not if it means we're opening a security hole. If a user does not want to open the port to the public, it should not be required to get the NLBService construct working. 

I think this needs more investigation to determine if something is missing/is a bug in the NetworkLoadBalancedService construct, or if this is the only solution.

iamhopaul123: I agree that we should not open a port (e.g., 80) to allow ingress traffic for every IP address. Using the console or aws CLI, we can obtain the IPv4 CIDR for each subnets for the NLB, then only allow ingress traffic for port 80 (or whatever port the listener listening to) for those IPv4 CIDRs so that the security problem can be solved.

However, it seems like CDK doesn't allow you to obtain those subnet CIDRs for the NLB. Essentially because the CIDRs can be known only in deploy stage. I would say in this example we can just leave it open to all IP, and in the README for ECS pattern on the NLB construct part, we should notify users that they can do it manually (setting ingress traffic rules for listener port, health check port) after their constructs are deployed.

typescript/ecs/ecs-network-load-balanced-service/index.ts
```diff
@@ -26,8 +31,9 @@ class BonjourECS extends cdk.Stack {
       }
     });
 
-    // Output the DNS where you can access your service
-    new cdk.CfnOutput(this, 'LoadBalancerDNS', { value: ecsService.loadBalancer.loadBalancerDnsName });
```
pkandasamy91: Why are you removing this?

typescript/ecs/fargate-application-load-balanced-service/index.ts
```diff
@@ -13,15 +13,12 @@ class BonjourFargate extends cdk.Stack {
...
-    const fargateService = new ecs_patterns.NetworkLoadBalancedFargateService(this, "FargateService", {
+    new ecs_patterns.ApplicationLoadBalancedFargateService(this, "FargateService", {
       cluster,
       taskImageOptions: {
         image: ecs.ContainerImage.fromRegistry("amazon/amazon-ecs-sample"),
       },
     });
-
-    // Output the DNS where you can access your service
-    new cdk.CfnOutput(this, 'LoadBalancerDNS', { value: fargateService.loadBalancer.loadBalancerDnsName });
```
pkandasamy91: We would want to keep this

typescript/ecs/ecs-network-load-balanced-service/index.ts
iamhopaul123: It will contain duplicated value. Since the higher level construct already output the same.

typescript/ecs/fargate-application-load-balanced-service/index.ts
iamhopaul123: the same reason as the previous one.

pkandasamy91:
@iamhopaul123 can you update the README?

iamhopaul123:
> Please update the readme as well

Nice catch. Fixed it!

PR #209, opened 2019-12-20
feat: cdk iam policy role generator

srihariph:
Fixes # 

---
This CDK Project helps create IAM Managed Policies and IAM Roles using JSON Configuration 


NGL321:
Hi @srihariph,

Thank you for taking the time to contribute!
This is a really interesting construct, but unfortunately, it isn't really what we are looking for in examples. We really want to see straightforward examples of very common patterns that people are likely going to need to reference when writing their code.

That being said, I dont want this to go to waste, so I suggest publishing it as its own 3rd party construct, and sharing it to our [Awesome CDK repo](https://github.com/eladb/awesome-cdk)!

PR #224, opened 2020-01-10
feat: Add EKS cluster with VPC and self managed ASG example

andskli:
Adds a typescript example EKS cluster with a VPC & auto-scaling group.


typescript/eks/cluster/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const vpc = new ec2.Vpc(this, 'EKSVpc');  // Create a new VPC for our cluster
+    
+    // IAM role for our EC2 worker nodes
+    const workerRole = new iam.Role(this, 'EKSWorkerRole', {
+      assumedBy: new iam.ServicePrincipal('ec2.amazonaws.com')
+    });
+
+    const eksCluster = new eks.Cluster(this, 'Cluster', {
+      clusterName: id,
```
eladb: It's an anti-pattern to specify a cluster name because any replacement-requiring updates will fail on it.

```suggestion
```

PR #240, opened 2020-02-05
feat: add typescript load balanced fargate service using shared load balancer

bvtujo:
This PR adds an example of how to create a new Fargate service using an existing load balancer and listener. This example, when all fields are filled in in the properties struct, will:

1) import a load balancer, listener, vpc, cluster, and security group for the load balancer
2) create a new service
3) allow security group ingress and egress from the load balancer to the new service
4) create a new target group using the service and port mappings
5) create a new listener rule on the imported listener to route traffic to the target group.

I originally produced this as a reply to a comment on [CDK PR 5461](https://github.com/aws/aws-cdk/pull/5461) and was asked to add it to the examples repository. 

---


bvtujo:
Would love some guidance here on how to handle an example which requires the existence of another stack with fully concrete (non-tokenized) resources to import. This use case comes directly from customer feedback.

peterwoodworth (dismissed):
Sorry for the wait and thank you for the contribution. I understand that this example requires the use of an existing stack, however we don't accept examples that don't run right out of the box. We could work on this example to create the already existing stack you require, but I understand that it's been a while and you might not be invested in this anymore.  Let me know if you're interested on getting this example up.

PR #278, opened 2020-04-23
fix: create a package.json inside of api-cors-lambda-crud-dynamodb's src folder

### LINKED ISSUE #276: typescript - api-cors-lambda-crud-dynamodb  -missing uuid in package.json
### What is the problem?
Error message `Error: Cannot find module 'uuid/v4` when testing `create` function.

### Reproduction Steps
1. deploy the stack as per instructions
2. go to the AWS Console and test `createItem` function with the following test event:

```
{
  "body": "{\"data\":{\"name\":\"foo\"}}"
}
```

3. The result will be that the task timed out. Go to Cloudwatch and view the log for the function. It will complain this: `"Error: Cannot find module 'uuid/v4`

### Environment

  - **CDK CLI Version:** 1.34.1
  - **Example:** api-cors-lambda-crud-dynamodb
  - **Example Version:** createItem function
  - **OS:**  OSX Catalina
  - **Language:** Typescript

### Other information 
The solution requires the uuid package being present inside a `node_modules` of the `src` folder. Currently there is no package.json inside the src folder.

nazreen:
I ran the following commands:

1. inside `src` folder, I ran `npm init`
2. then, in the same folder, I ran `npm install uuid`
3. I updated the README to reflect the new required step of running `npm i` inside the `src` folder

Fixes #276 
---


nazreen:
the build failure is because there is no build script for the package.json under `src`. however, the project root already has typescript in it that covers the `src`folder.

@NGL321  what's the best move forward? would there still need to be a dummy build script?

nazreen:
another error message after creating the dummy build script: `--app is required either in command-line, in cdk.json or in ~/.cdk.json`

I think the solution has to do with using the root directory to install node_modules in the src directory (install npm module in subdirectory)

NGL321:
@nazreen I apologize again for the delay in response,

As I mentioned in the [related issue](https://github.com/aws-samples/aws-cdk-examples/issues/276#issuecomment-625569581), it seems that the issue actually has to do with import method and is a breaking change introduced by the UUID package.

If you would like to edit this PR to the proper fix I would be happy to merge it! Alternatively feel free to put in a new PR with the fix and @ me in it.

Cheers!
😷

nazreen:
@NGL321 ah I see! thank you so much for pointing that out, I wasn't aware of that change for uuid. I will submit a new PR :)

nazreen:
note: I will be creating another PR for this issue, which will be cleaner.

nazreen:
after trying to create a dummy `cdk.json` inside of `src`, and setting the `app` value to `node ../index`, the build step tries to look for `src/src`

it seems like this is a check triggered by the existence of a `package.json` because it assumes that every `package.json` would be in the root and lambdas would not have `package.json`'s in their `src` folders.

nsutcliffe-speedline:
I ran into this issue too and I'm really curious what Amazon's recommended approach will be.  Developers using Node.js with Lambda are going to want a good solution for including libraries from NPM.

Having package.json in the root directory and a different one in the src directory could probably work but it seems a bit awkward and unusual to me.

Is a better way to structure this type of project (a service with several non-trivial Lambda functions using Node.js) to use the [aws-lambda-nodejs library](https://github.com/aws/aws-cdk/tree/master/packages/%40aws-cdk/aws-lambda-nodejs)?  That library has explicit support for bundling third-party libraries.  It looks to me like you would have the option with that library to have one pacakge.json or to have several (managed by something like Lerna).

sachin10101998:
@nazreen  how do we fix this?  @NGL321

chrischenyc:
I think using the experimental [@aws-cdk/aws-lambda-nodejs](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-lambda-nodejs-readme.html) is the way to go, works for me:

https://github.com/chrischenyc/url-shortener-cdk-typescript

ryparker:
Because this PR branch is protected I created a new PR branch with the necessary fixes: https://github.com/aws-samples/aws-cdk-examples/pull/468

Once merged this should close this PR and referenced issue.

ryparker:
Because this PR branch is protected I created a new PR branch with the necessary fixes.

Closing in favor of https://github.com/aws-samples/aws-cdk-examples/pull/468

PR #282, opened 2020-05-01
feat: Add CDK Example ec2 in vpc with ssm

kristianpaul:
With this pull request I want to  publish a CDK Python example that deploys an EC2 instance taking advantage of AWS services like System Manager, Amazon Linux 2 AMI and  CDK Assets to package scripts that provision instance as needed.

* Own VPC with public subnet (following AWS Defaults for new accounts)
* Based on latest Amazon Linux 2
* System Manager replaces SSH (Remote session available trough the AWS Console or the AWS CLI.)
* Userdata executed from script in S3 bucket (`configuration.sh`).

---


python/ec2/instance/configure.sh
```diff
@@ -0,0 +1 @@
+#!/bin/sh
```
rix0rrr, line 1: Could do with a comment in this file explaining what it's for.

python/ec2/instance/setup.py
```diff
@@ -0,0 +1,45 @@
+import setuptools
```
rix0rrr, line 1: I guess this setup.py doesn't add much, does it?

python/ec2/instance/app.py
```diff
@@ -0,0 +1,63 @@
...
+        asset = Asset(self, "Asset", path=os.path.join(dirname, "configure.sh"))
+        local_path = instance.user_data.add_s3_download_command(
+            bucket=asset.bucket,
+            bucket_key=asset.s3_object_key
+        )
+
+        # Userdata executes script from S3
+        instance.user_data.add_execute_file_command(
+            file_path=local_path,
+            arguments="--verbose -y"
```
rix0rrr, line 56: ?

python/ec2/instance/app.py
```diff
@@ -0,0 +1,63 @@
...
+            )
+
+        # Instance Role and SSM Managed Policy
+        role = iam.Role(self, "InstanceSSM", assumed_by=iam.ServicePrincipal("ec2.amazonaws.com"))
+
+        role.add_managed_policy(iam.ManagedPolicy.from_aws_managed_policy_name("service-role/AmazonEC2RoleforSSM"))
+
+        # Instance
+        instance = ec2.Instance(self, "Instance",
+        instance_type=ec2.InstanceType("t3.nano"),
```
rix0rrr, line 40: Needs indentation here.

python/ec2/instance/app.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class EC2InstanceStack(core.Stack):
+
+    def __init__(self, scope: core.Construct, id: str, **kwargs) -> None:
+        super().__init__(scope, id, **kwargs)
+
+        # VPC
+        vpc = ec2.Vpc(self, "VPC",
+        nat_gateways=0,
```
rix0rrr, line 21: Needs indentation here.

python/ec2/instance/app.py
kristianpaul: good catch, fixing

PR #283, opened 2020-05-04
feat:  Add a new CDK python example for Cloudwatch event rule and AWS Backup  

Mason117:
With this pull request I want to publish a CDK Python example demonstrates:

 * Create a new VPC with S3 endpoint
 * Two Instances, a web server and a bastion host
 * A cloudwatch event rule to stop instances at UTC 15pm every day
 * A backup vault and a backup rule to protect resources 
 * Show output example

---


Mason117:
@MrArnoldPalmer  
Hi, I have changed my code as @richardhboyd said, what shoud I do now? Make a new PR?  Besides,  I have tested my code in a real AWS environment, but the Build Pull Requests here was failed, error information said something wrong with the account certificate.  Cloud you please have a check?

MrArnoldPalmer:
The build is failing with `[Error at /ec2-cloudwatch] Need to perform AWS calls for account unkonw, but no credentials have been configured.
Found errors` which is caused by [this code here](https://github.com/aws-samples/aws-cdk-examples/pull/283/files#diff-51e0be93d181332cacdbad2c8b2ff7bcR12-R13).

CDK automatically resolves to outputting "unknown" in the CFN template if an account number isn't specified during synth. You shouldn't need to handle this. For the purpose of the example, I'd recommend removing this logic so users can dictate the account/environment the same way across all of the examples.

Mason117:
@MrArnoldPalmer 
Hi, Valine, I have changed my code as your suggested and passed all checks. Could you please spend some time to review my code ?

python/ec2-cloudwatch/source.bat
```diff
@@ -0,0 +1,13 @@
+@echo off
```
MrArnoldPalmer, line 1: Lets remove this file. Generally I think it will confuse users more than help. Our examples assume that users have their tooling setup however they like while this assumes the use of virtualenv.

PR #302, opened 2020-06-17
feat: add python emr cluster example

fmcmac:
Add a python example for an EMR cluster as one does not currently exist.
Cluster contains a single step


fmcmac:
@NGL321 I have updated the maps into CDK interfaces.

NGL321 (approved):
Sorry for the long delay. Reviewed, looks like a great addition!

🚀 😸

PR #306, opened 2020-06-22
fix(python/rds): upgrade to new API

RomainMuller:
Upgrades to the latest API from the RDS package.

---

### DIFF
--- a/scripts/build-java.sh
+++ b/scripts/build-java.sh
@@ -16,7 +16,7 @@ for pomFile in $(find $scriptdir/../java -name pom.xml); do
         cd $(dirname $pomFile)
         if [[ -f DO_NOT_AUTOTEST ]]; then exit 0; fi
 
-        mvn compile test
+        mvn -q compile test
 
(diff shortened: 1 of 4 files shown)


scripts/build-java.sh
```diff
@@ -16,7 +16,7 @@ for pomFile in $(find $scriptdir/../java -name pom.xml); do
         cd $(dirname $pomFile)
         if [[ -f DO_NOT_AUTOTEST ]]; then exit 0; fi
 
-        mvn compile test
+        mvn -q compile test
```
njlynch, line 19: What's the motivation here? The output isn't _that_ noisy, and part of me always wants just a bit more info when diagnosing a problem. That being said, this is the first time I'm looking at this build logs, so happy to defer to your judgment.

RomainMuller: Well I just had to spend ~10 minutes scrolling down past maven output (because of the verbosity of it's downloads) in Github Actions to find the test that *really* failed... It was not pleasant.

PR #317, opened 2020-07-25
fix: typo in python fargate load balancer ecs example

damo-da:
kwargs is accepted as ** but passed as *. Fixed typo.


python/ecs/fargate-load-balanced-service/app.py
```diff
@@ -9,7 +9,7 @@
 class BonjourFargate(core.Stack):
 
     def __init__(self, scope: core.Construct, id: str, **kwargs) -> None:
-        super().__init__(scope, id, *kwargs)
+        super().__init__(scope, id, **kwargs)
```
shivlaks, line 12:
```suggestion
        super().__init__(scope, id, **kwargs)
```

shivlaks (dismissed):
seems like the recent changes to rds also need to be updated - i believe we've addressed it and it might just need a merge / update

ryparker (approved):

PR #321, opened 2020-07-31
fix: Added version to eks.Cluster with 1.17 as version to be compatible with cdk v1.55.0

dnmach:
…th cdk v1.55.0

Fixes # 

---


NGL321 (dismissed):
Thanks for the fix, and sorry for the delay in review!

😸 😷

typescript/eks/cluster/index.ts
```diff
@@ -16,6 +16,7 @@ class EKSCluster extends cdk.Stack {
     });
 
     const eksCluster = new eks.Cluster(this, 'Cluster', {
+      version: eks.KubernetesVersion.V1_17,
```
shivlaks, line 19: there's a version key on line 22 that should have it's value updated instead of introducing a duplicate key

shivlaks (changes requested):

typescript/eks/cluster/index.ts
```diff
@@ -16,6 +16,7 @@ class EKSCluster extends cdk.Stack {
...
       vpc: vpc,
       defaultCapacity: 0,  // we want to manage capacity our selves
       version: eks.KubernetesVersion.V1_16,
```
ryparker, line 22:
```suggestion
```
This is a duplicate key.

ryparker (changes requested):

ryparker:
Changes already merged.

PR #336, opened 2020-09-23
feat(Typescript): StepFunctions CRON example

michchan:
Created an example of how to create sequential control CRON with lambda and StepFunctions.
It is useful when there are tasks that requires another resource to be active, which is dynamically created (e.g. table creation).

For example, I want to create a time-series application with CDK. The best approach is to dynamically create a table for each period (e.g. 1 day, 1 week or 1 month etc.). Then the CRON process each time will be like:
1. Check if there is a table for the current period. If it does not exist, create one.
2. Wait for the status of the table created to be `active`.
3. Put new records.


peterwoodworth (dismissed):
Hi, thank you for contributing, we appreciate all community contributions. That said, we're concerned that this doesn't add additional value beyond [this example](https://github.com/aws-samples/aws-cdk-examples/tree/master/typescript/stepfunctions-job-poller). We do however like your code, and it raises the bar for the examples. 

There are a couple paths forward from here, either you can convert this new example to be a new revision of the previous one, or we can figure out some additional value to add to this contribution before merging. I recommend using this to update the old example so that we can keep the code there current and high quality.

Please let us know which path, if either, you would like to follow through with and we can help you figure out what will be required.

michchan:
> Hi, thank you for contributing, we appreciate all community contributions. That said, we're concerned that this doesn't add additional value beyond [this example](https://github.com/aws-samples/aws-cdk-examples/tree/master/typescript/stepfunctions-job-poller). We do however like your code, and it raises the bar for the examples.
> 
> There are a couple paths forward from here, either you can convert this new example to be a new revision of the previous one, or we can figure out some additional value to add to this contribution before merging. I recommend using this to update the old example so that we can keep the code there current and high quality.
> 
> Please let us know which path, if either, you would like to follow through with and we can help you figure out what will be required.

Thanks @peterwoodworth. 
I do agree that updating the old one would be the better path.
Please help review the latest change.

ryparker:
Thanks for the update! Will test on our end and get back to you.

ryparker (dismissed):

ryparker:
Made some fixes to get it working again. Looks good. Thanks for contributing!

ryparker:
@Mergifyio refresh

ryparker (approved):

michchan:
> Made some fixes to get it working again. Looks good. Thanks for contributing!

Welcome and thanks for the fix @ryparker!

PR #364, opened 2020-11-11
feat: add AWS transfer with custom IdP example for python

darren-reddick:
This PR is to add an example of using creating a Custom IdP for the Transfer for SFTP service.

The authentication is handled by a Lambda-backed API Gateway which authenticates the username:password against AWS Secrets Manager secrets.

I previously have created a Terraform module to achieve this which functions but can be temperamental with updates etc. The CDK seems to handle the dependencies between the API Gateway components much better.

---


indrora:
Thank you for your contribution. Unfortunately, this PR is very old, has had no movement, and predates the launch of CDK v2. 

I'm going to close this PR for now. If you update it to use the CDKv2, please open a new PR.

PR #379, opened 2020-12-31
chore: Added notes about broken examples

iliapolo:
This [issue](https://github.com/aws/aws-cdk/issues/12291) broke several examples. Added a note about this breakage with a link to possible workarounds.

Once we release a patch, the builds should be fixed and we can remove those disclaimers. 

---

### DIFF
--- a/csharp/static-site/README.md
+++ b/csharp/static-site/README.md
@@ -18,6 +18,8 @@ This example creates the infrastructure for a static site, which uses an S3 buck
 
 The site redirects from HTTP to HTTPS, using a CloudFront distribution, Route53 alias record, and ACM certificate.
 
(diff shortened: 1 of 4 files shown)


iliapolo:
The build will fail because of the mentioned issue, we would need to manually force merge to get the docs updated.

NGL321 (approved):
Thank you for the update!

iliapolo:
@NGL321 We can close this one right? we already have a fix in the latest available version

NGL321:
> @NGL321 We can close this one right? we already have a fix in the latest available version

Looks like you are right, I need to check into why automerge didnt work initially to merge this PR.

PR #384, opened 2021-01-21
feat(Typescript): Microsoft Active Directory + FSx for Windows File Server example

javydekoning:
Added examples in the Typescript language for creating an FSx for Windows File Server filesystem which is joined to an Managed Microsoft Active Directory in AWS.

---


ryparker:
Thanks for contributing this. Testing this now.

ryparker:
I was able to build the stack however I'd like to add something that demos how to use the AD rather than just how to standup the AD.

javydekoning:
Hi Ryan, any suggestions? For example,I could ad join an FSx filesystem?

ryparker (approved):
I think that this example is good to merge as is. We can add other features to the example as needed.

PR #395, opened 2021-02-28
feat: TypeScript example for rotating ElastiCache Redis Auth Token with Secrets Manager #393

### LINKED ISSUE #393: Typescript example of rotating credentials for a non-supported DB type
### General Information

* [X] :wave: I may be able to implement this feature request

* [ ] :warning: This feature might incur a breaking change 

### Description

Looking at the documentation, it looks like there are only cli examples for rotating credentials for non-supported resource types from Secrets Manager [here](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-create-generic-template.html)
 
### Proposed Solution

The proposed solution will deploy an ElastiCache Redis cluster with Auth token enabled; the Auth token will be generated by AWS Secrets Manager and rotated with a custom Lambda function.  Another Lambda function will act as a mock application that will attempt to connect to the ElastiCache Redis cluster both before and after token rotation.

### Environment

  - **CDK CLI Version:** Latest (1.89.0)
  - **Example:** (proposed) secret-manager-rotation-elasticache-redis
  - **Example Version:** (proposal)
  - **OS:** All
  - **Language:** TypeScript

### Other information 

https://docs.aws.amazon.com/secretsmanager/latest/userguide/enable-rotation-other.html
https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html

its-mirus-lu:
This code demonstrates sample infrastructure to rotate an Auth token for ElastiCache redis.  The example shows how to implement a rotation Lambda and the infrastructure related to it.

Fixes #393 

---


typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    secret.grantRead(fn);
+
+    fn.grantInvoke(new iam.ServicePrincipal('secretsmanager.amazonaws.com'))
+
+
+  }
+}
+
+const app = new cdk.App();
```
ryparker, line 179:
```suggestion
const app = new App();
```

Upgrade to CDK v2

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      license: 'MIT License'
+    });
+
+    const fn = new lambda.Function(this, 'function', {
+      runtime: lambda.Runtime.PYTHON_3_7,
+      handler: 'lambda_handler.lambda_handler',
+      code: lambda.Code.fromAsset(path.join(__dirname, 'lambda')),
+      layers: [redisPyLayer],
+      role: rotatorRole,
+      timeout: cdk.Duration.seconds(30),
```
ryparker, line 153:
```suggestion
      timeout: Duration.seconds(30),
```
Upgrade to CDK v2

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+import ec2 = require('@aws-cdk/aws-ec2');
+import iam = require('@aws-cdk/aws-iam');
+import elasticache = require('@aws-cdk/aws-elasticache');
+import lambda = require('@aws-cdk/aws-lambda');
+import secretsmanager = require('@aws-cdk/aws-secretsmanager')
+import path = require('path');
+
+
+export class SecretsManagerCustomRotationStack extends cdk.Stack {
+  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
```
ryparker, line 11:
```suggestion
import { App, Stack, StackProps, Duration } from 'aws-cdk-lib';
import { aws_ec2 as ec2 } from 'aws-cdk-lib'; 
import { aws_iam as iam } from 'aws-cdk-lib'; 
import { aws_elasticache as elasticache } from 'aws-cdk-lib'; 
import { aws_lambda as lambda } from 'aws-cdk-lib'; 
import { aws_secretsmanager as secretsmanager } from 'aws-cdk-lib'; 
import path = require('path');


export class SecretsManagerCustomRotationStack extends Stack {
  constructor(scope: cdk.App, id: string, props?: StackProps) {
```

Upgrade to CDK v2

typescript/secrets-manager-secret-rotation/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "@aws-cdk/aws-ec2": "*",
+    "@aws-cdk/aws-elasticache": "*",
+    "@aws-cdk/aws-iam": "*",
+    "@aws-cdk/aws-lambda": "*",
+    "@aws-cdk/aws-secretsmanager": "*",
+    "@aws-cdk/aws-sns": "*",
+    "@aws-cdk/aws-sns-subscriptions": "*",
+    "@aws-cdk/aws-sqs": "*",
+    "@aws-cdk/core": "*"
+  }
```
ryparker, line 37:
```suggestion
  "dependencies": {
    "aws-cdk-lib": "*",
    "constructs": "*",
  }
```

Would you mind upgrading this to CDK v2 for consistency with the other projects we have? I've added code suggestions with all the necessary code to upgrade. It's possible I may have missed something though but let me know if you have trouble.

By the way you'll need to update the lockfile once these suggestions are accepted. Run `npm install` and commit the updated lockfile changes.

ryparker (changes requested):
Thanks for contributing! I have a few suggestions before we merge this. 

By the way in your png there is a secret being logged. Just wanted to let you know in case the screenshot is from an existing production project you have.

See `typescript/secrets-manager-secret-rotation/img/output.png`

typescript/secrets-manager-secret-rotation/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.ts
+!*.d.ts
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
```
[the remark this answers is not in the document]
its-mirus-lu: Good catch, this isn't needed since npm will fallback to .gitignore if .npmignore isn't present. Removing npmignore would simplify things.

its-mirus-lu:
> Thanks for contributing! I have a few suggestions before we merge this.
> 
> By the way in your png there is a secret being logged. Just wanted to let you know in case the screenshot is from an existing production project you have.
> 
> See `typescript/secrets-manager-secret-rotation/img/output.png`

Thanks for catching that. It was part of a PoC stack that has long been torn down, but yes, that should be obfuscated in the image and it will be.

typescript/secrets-manager-secret-rotation/package.json
its-mirus-lu: Certainly, I'll upgrade this to v2 and commit the changes.

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      description: 'SecurityGroup for rotator function'
+    });
+
+    rotatorSecurityGroup.addIngressRule(ec2.Peer.anyIpv4(), ec2.Port.allTraffic(), 'All port inbound')
+
+    let isolatedSubnets: string[] = []
+
+    vpc.isolatedSubnets.forEach(function(value){
+      isolatedSubnets.push(value.subnetId)
+    });
```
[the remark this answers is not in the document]
its-mirus-lu: This is more elegant

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+          cidrMask: 24,
+          name: 'Private',
+          subnetType: ec2.SubnetType.PRIVATE,
+        },
+        {
+          cidrMask: 24,
+          name: 'Public',
+          subnetType: ec2.SubnetType.PUBLIC,
+        },
+      ]
```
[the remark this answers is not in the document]
its-mirus-lu: Good catch, this slipped by me.

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      isolatedSubnets.push(value.subnetId)
+    });
+
+    const ecSubnetGroup = new elasticache.CfnSubnetGroup(this, 'ElastiCacheSubnetGroup', {
+      description: 'Elasticache Subnet Group',
+      subnetIds: isolatedSubnets
+    });
+
+
+
```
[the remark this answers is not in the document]
its-mirus-lu: done

typescript/secrets-manager-secret-rotation/index.ts
its-mirus-lu: Roger that

typescript/secrets-manager-secret-rotation/index.ts
its-mirus-lu: Done

typescript/secrets-manager-secret-rotation/index.ts
its-mirus-lu: Updated

typescript/secrets-manager-secret-rotation/package.json
its-mirus-lu: Done and done!

its-mirus-lu:
@ryparker, I just pushed an omnibus commit that updates the example to CDK v2 as well as some cleanup and obfuscating that image in the readme.

Let me know if you guys are hiring ;)

typescript/secrets-manager-secret-rotation/lambda/lambda_handler.py
```diff
@@ -0,0 +1,305 @@
...
+            % (arn, token)
+        )
+
+
+def set_secret(service_client, arn, token):
+    #   {
+    #   "Step": "Set",
+    #   "SecretId": "arn:aws:secretsmanager:us-east-1:323144477050:secret:RedisAuthAF1E8ACD-ahzrCUonZBP7-PzbBrY",
+    #   "ClientRequestToken": "ca87e860-e419-4cce-9e62-b1d2c7e1cede"
+    # }
```
ryparker, line 148: Not sure what this is? Would you mind adding an explanation/remove?

typescript/secrets-manager-secret-rotation/lambda/lambda_handler.py
```diff
@@ -0,0 +1,262 @@
...
+        service_client (client): The secrets manager service client
+
+        arn (string): The secret ARN or other identifier
+
+        token (string): The ClientRequestToken associated with the secret version
+
+    """
+    # This is where the secret should be set in the service
+
+    # service_client = boto3.client("secretsmanager")
```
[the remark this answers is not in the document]
ryparker: Any update^?

typescript/secrets-manager-secret-rotation/lambda/lambda_handler.py
```diff
@@ -0,0 +1,262 @@
...
+    arn = event['SecretId']
+    token = event['ClientRequestToken']
+    step = event['Step']
+
+
+    # Setup the client
+    service_client = boto3.client('secretsmanager', endpoint_url=os.environ['SECRETS_MANAGER_ENDPOINT'])
+
+    # ***
+
```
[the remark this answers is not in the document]
ryparker: ^

typescript/secrets-manager-secret-rotation/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      cacheSubnetGroupName: ecSubnetGroup.ref,
+      engine: "Redis",
+      engineVersion: '6.x',
+      numNodeGroups: 1,
+      replicasPerNodeGroup: 1,
+      securityGroupIds: [ecSecurityGroup.securityGroupId],
+      transitEncryptionEnabled: true,
+      authToken: secret.secretValueFromJson('authToken').toString()
+    })
+
```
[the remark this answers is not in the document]
ryparker: Does this^ sound ok to you? I can offer some suggested replacements if necessary.

ryparker (changes requested):
Thank you for making those changes. It looks like there are a few review comments that may have been accidentally skipped. I put a comment on each of them in this review for you to consider. Let me know if you have questions. 🙏🏻 

AWS is always hiring, check out some of our [CDK opportunities](https://www.amazon.jobs/en/search?base_query=%22CDK%22&loc_query=&latitude=&longitude=&loc_group_id=&invalid_location=false&country=&city=&region=&county=). 😄

typescript/secrets-manager-secret-rotation/lambda/lambda_handler.py
its-mirus-lu: I'll remove this. This was a note that I left for debugging.

its-mirus-lu:
> Thank you for making those changes. It looks like there are a few review comments that may have been accidentally skipped. I put a comment on each of them in this review for you to consider. Let me know if you have questions. 🙏🏻
> 
> AWS is always hiring, check out some of our [CDK opportunities](https://www.amazon.jobs/en/search?base_query=%22CDK%22&loc_query=&latitude=&longitude=&loc_group_id=&invalid_location=false&country=&city=&region=&county=). 😄

I've gone ahead and cleaned up the remaining scruff comments that you found (good eye). You also stumbled onto something in the test stage of the rotation flow -- I've added a wait loop before attempting to connect to the redis cluster and also raised an error in the actual event that the client connection failed.  Have a look and let me know what you think.

... in the meantime I'll have a look at the cdk job link you sent ;)

its-mirus-lu:
@ryparker I wanted to follow up on this discussion as well https://github.com/aws-samples/aws-cdk-examples/pull/395#discussion_r811334253

Were you pointing to decrypting the secret as plaintext to be passed to the elasticache boto call later in the method?  Is there a better way to do this? I was under the impression that the method call to update the replication group required a plaintext auth token?

ryparker:
> 

Sorry for the confusion I actually was referring to using .connections instead of creating security groups. Here's the original comment I was pointing(^) to:

> We usually recommend using the .connections.* API instead of configuring security groups, it'll also reduce the code here.
> 
> i.g. The following should remove the need for any security groups in this code.
> 
> ```ts
> ecClusterReplicationGroup.connections.allowFromAnyIpv4(6379, 'Redis ingress 6379');
> ...
> fn.connections.allowDefaultPortFrom(ecClusterReplicationGroup);
> fn.connections.allowFromAnyIpv4(ec2.Port.allTraffic(), 'All port inbound')
> ```

its-mirus-lu:
> > 
> 
> Sorry for the confusion I actually was referring to using .connections instead of creating security groups. Here's the original comment I was pointing(^) to:
> 
> > We usually recommend using the .connections.* API instead of configuring security groups, it'll also reduce the code here.
> > i.g. The following should remove the need for any security groups in this code.
> > ```ts
> > ecClusterReplicationGroup.connections.allowFromAnyIpv4(6379, 'Redis ingress 6379');
> > ...
> > fn.connections.allowDefaultPortFrom(ecClusterReplicationGroup);
> > fn.connections.allowFromAnyIpv4(ec2.Port.allTraffic(), 'All port inbound')
> > ```
> 
> _You can see the [original comment here](https://github.com/aws-samples/aws-cdk-examples/pull/395/files#r807304324)_

Ooh this is new to me. It does abstract things quite a bit. There is one issue though, the aws_elasticache library only has an L1 construct for [cache clusters](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticache.CfnCacheCluster.html) and there isn't a connections method or similar abstraction method for CfnCacheCluster.

I can definitely update the Lambda function to use the connections method, but it looks like we'll still have to explicitly define security groups for the CfnCacheCluster.

ryparker:
> 

Ah in this case don't worry about it. IMO we should keep it consistent and once the L2 construct is released we can update it with the latest syntax. Thanks for the explanation.

ryparker:
Just finished testing this without failures 💯 . However I am a little concerned about us keeping the redis_py module zip in this repo. It'll be difficult for us to maintain its versioning and it could easily be maliciously modified. What if we instead installed it at runtime via a requirements.txt? We should be able to utilize the CDK's [aws-lambda-python module](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-lambda-python-readme.html) to accomplish this.

If you think this is a reasonable approach let me know. I can help implement this as well, i've already started implementing this on my private branch. If you're ok with this update i'll open a PR to this branch for you to review once I have it working.

its-mirus-lu:
> Just finished testing this without failures 💯 . However I am a little concerned about us keeping the redis_py module zip in this repo. It'll be difficult for us to maintain its versioning and it could easily be maliciously modified. What if we instead installed it at runtime via a requirements.txt? We should be able to utilize the CDK's [aws-lambda-python module](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-lambda-python-readme.html) to accomplish this.
> 
> If you think this is a reasonable approach let me know. I can help implement this as well, i've already started implementing this on my private branch. If you're ok with this update i'll open a PR to this branch for you to review once I have it working.

That's pretty neat. Yes, by all means, have at it if you've started implementing it!

ryparker:
Done! Check it out here: https://github.com/its-mirus-lu/aws-cdk-examples/pull/1

its-mirus-lu:
heh... @ryparker it looks like we're breaking new ground (and builds) with the lambda python library... the Github runner(s) will need to be configured to have the Docker daemon running to run the bundling process in the `PythonLayerVersion` construct...

from the build log: `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
docker exited with status 1`

ryparker:
> 

This is an interesting problem indeed. I'm looking into a few alternatives. I should have an update for you tomorrow.

ryparker:
Hey just wanted to give you an update as promised.

It doesn't look like we can simply expose the Docker Daemon Socket or use the Docker in Docker container in our CI. I think the solution i'm leaning towards is to write a shell script that installs the python requirements.txt and zips them for use in a CDK Lambda Layer. This is similar to what you had originally except we fetch the deps at runtime and are able to manage versioning via requirements.txt. As the code stands now it is actually in working condition, however we're limited by CI.

Another option that I considered is to skip the building of this project and instead have some very comprehensive tests. I'm not a huge fan of this option because there is always the possibility of the tests not catching everything. This will also add one-off logic to our CI.

 I'll need a little more time to find a permanent solution to this. I'll follow up here next week. 😄

its-mirus-lu:
This was mistakenly closed when I tried to a git filter-branch command to amend a couple of commits I did with a wrong github account

PR #401, opened 2021-03-23
fix: Loading existing image from aws_ecr does not work

nuno-seco:
The existing code produced invalid cloudformation formation code. This code change made it work for me on latest versions of cdk. I do not see any existing issue for this


ryparker (dismissed):
Tested 👍🏻

ryparker (approved):

PR #405, opened 2021-03-31
feat(go): EKS Cluster

iliapolo:
Adding the golang equivalent of the typescript EKS cluster example.

Related to https://github.com/aws/jsii/issues/2676

---


go/eks-cluster/main.go
```diff
@@ -0,0 +1,48 @@
...
+		MaxCapacity: jsii.Number(10),
+		InstanceType: ec2.NewInstanceType(jsii.String("t3.medium")),
+		MachineImage: eks.NewEksOptimizedImage(&eks.EksOptimizedImageProps{
+			KubernetesVersion: jsii.String("1.14"),
+			NodeType: eks.NodeType_STANDARD,
+		}),
+		UpdateType: autoscaling.UpdateType_ROLLING_UPDATE,
+	});
+
+	eksCluster.ConnectAutoScalingGroupCapacity(onDemandASG, &eks.AutoScalingGroupOptions{});
```
iliapolo, line 45: For some reason I cannot use `nil` here:

```console
❯ cdk synth                                                                       [01:12:48]
panic: "Got 'undefined' for non-optional instance of {\"docs\":{\"summary\":\"options for adding auto scaling groups, like customizing the bootstrap script.\"},\"name\":\"options\",\"type\":{\"fqn\":\"monocdk.aws_eks.AutoScalingGroupOptions\"}}"

goroutine 1 [running]:
github.com/aws/jsii-runtime-go.InvokeVoid(0x1698c00, 0xc0004b4d20, 0x16ba4c6, 0x1f, 0xc00006fe48, 0x2, 0x2)
```

iliapolo: Scratch that. The problem is that the argument is indeed non optional, even though all its properties are. So it makes sense `nil` is not an allowed value.

eladb (approved):

PR #408, opened 2021-04-01
feat(go): lambda-cron

iliapolo:
Added the go version of [`lambda-cron`](https://github.com/aws-samples/aws-cdk-examples/tree/master/typescript/lambda-cron)
 
---


go/lambda-cron/lambda-cron.go
```diff
@@ -0,0 +1,89 @@
...
+	// The code that defines your stack goes here
+
+	code, err := ioutil.ReadFile("lambda-handler.py")
+
+	if err != nil {
+		panic(fmt.Sprintf("Failed reading handler code: %s", err))
+	}
+
+	lambdaFn := awslambda.NewFunction(stack, jsii.String("Singleton"), &awslambda.FunctionProps{
+		Code: awslambda.NewInlineCode(jsii.String(string(code))),
```
eladb, line 35: Why inline code and not an asset? Is that the same as the original sample?

iliapolo: Yeah its the same in the original. But happy to change, this will also show a direct asset usage we still don't have 👍

eladb: Yes, let's do it

iliapolo: Done. @eladb

eladb (approved):

Issue #195, opened 2019-11-29
api-cors-lambda-crud-dynamodb POST timeout


DavidQN:
### What is the problem?

I believe this could also fall into `Guidance`.

Specifically for: `typescript`/`api-cors-lambda-crud-dynamodb` example code, after building and deploying the stack. If you try to run the `POST` request you get a timeout (3 seconds). 

This is done using an example body such as:
```json
{
    "body":{
        "temp": "hello!!"
    }
}

``` 
- All other methods in this example code seem to work exactly as expected, except for `create`.
- If I run this locally and call the `create.js` from my machine, inserting into dynamodb seems to work fine.  So the code isn't the issue since the ENV variables are correct.  
- This lead me to believe that perhaps this has something to do with permissions on the create lambda? Since there is a timeout, which I believe occurs on the connection to the dynamodb table on line 23: `await db.put(params).promise();` since no error is ever outputted and just a timeout.
- I checked the cloudformation output, but they all seem to be correct:
    - Correct ServiceRole: "lambda.amazonaws.com"
![image](https://user-images.githubusercontent.com/23644019/69878806-eafa8300-1293-11ea-9cd4-5a6192311dbc.png)

    - Correct Policies for DynamoDb attached to said service role
![image](https://user-images.githubusercontent.com/23644019/69879293-0ade7680-1295-11ea-8134-7276d2e20e1d.png)

- All attached to the lambda in question
![image](https://user-images.githubusercontent.com/23644019/69879030-6c521580-1294-11ea-9c1f-fff74c9e489e.png)

### Reproduction Steps
- cd into `typescript/api-cors-lambda-crud-dynamodb/`
- npm install
- cdk deploy
- Go to Apigateway and in POST  /items place this body
```json
{
    "body":{
        "temp": "hello!!"
    }
}
``` 

### Verbose Log

- Response: 
![image](https://user-images.githubusercontent.com/23644019/69879634-f0f16380-1295-11ea-80f7-c14f553b3a5a.png)

### Environment

  - **CDK CLI Version: 1.16.2 ** 
  - **Example: api-cors-lambda-crud-dynamodb ** 
  - **Example Version:  ** 
  - **OS: Windows ** 
  - **Language: Typescript ** 

Exact same environment as the example code

rhboyd:
does every POST fail, or just the first?

DavidQN:
Every POST fails. This doesn't seem to be a lambda warmup issue if that is what you were getting at.

[...]

PR #468, opened 2021-07-08
fix(api-cors-lambda-crud-dynamodb example): Implement use of NodejsFunction, replaced deprecated API

### LINKED ISSUE #276: typescript - api-cors-lambda-crud-dynamodb  -missing uuid in package.json

ryparker:
## Summary

This PR changes the following:

- Implement use of NodejsFunction.
- Upgraded lambda runtime from Node 10 -> 14.
- Renamed `/src/` -> `/lambdas/`. This was a personal choice but I believe it makes more sense to label this as anything but `src` since it does not share the same node_modules as the root `package.json`.
- Created `/lambdas/package.json` to define modules required by the lambda functions.
- Added `"postinstall"` script to root package.json which installs the nested `/lambdas/package.json` dependencies after running `npm install` from the root.

Fixes #276, #468, #195
Would have pushed my fixes to #278 however the PR author's branch is protected. 

---


typescript/api-cors-lambda-crud-dynamodb/index.ts
```diff
@@ -14,62 +16,94 @@ export class ApiLambdaCrudDynamoDBStack extends cdk.Stack {
...
+       * DESTROY, cdk destroy will delete the table (even if it has data in it)
+       */
       removalPolicy: cdk.RemovalPolicy.DESTROY, // NOT recommended for production code
     });
 
-    const getOneLambda = new lambda.Function(this, 'getOneItemFunction', {
-      code: new lambda.AssetCode('src'),
-      handler: 'get-one.handler',
-      runtime: lambda.Runtime.NODEJS_10_X,
+    const getOneLambda = new NodejsFunction(this, 'getOneItemFunction', {
```
NGL321, line 27: If you are going to use direct component imports (ie. `import { NodejsFunction } from '...';` vs `import * as lambdanode from '...';` and `new NodejsFunction()` vs `new lambdanode.NodejsFunction()`) please do so for all other components (`lambda.Runtime`, `dynamodb.Table`, etc.)

NGL321: Just to clarify, I think importing components directly is a good approach for examples as it looks cleaner and is considered good practice for projects with few imports.

typescript/api-cors-lambda-crud-dynamodb/index.ts
```diff
@@ -14,62 +16,94 @@ export class ApiLambdaCrudDynamoDBStack extends cdk.Stack {
...
         PRIMARY_KEY: 'itemId'
       }
     });
 
-    const updateOne = new lambda.Function(this, 'updateItemFunction', {
-      code: new lambda.AssetCode('src'),
-      handler: 'update-one.handler',
-      runtime: lambda.Runtime.NODEJS_10_X,
+    const updateOne = new NodejsFunction(this, 'updateItemFunction', {
+      entry: join(__dirname, 'lambdas', 'update-one.ts'), // accepts .js, .jsx, .ts and .tsx files
```
NGL321, line 76: Imo these comments only need to appear on the first occurrence (unless you have strong feelings to the contrary).

typescript/api-cors-lambda-crud-dynamodb/lambdas/create.ts
```diff
@@ -1,13 +1,15 @@
-const AWS = require('aws-sdk');
-const db = new AWS.DynamoDB.DocumentClient();
-const uuidv4 = require('uuid/v4');
+import * as AWS from 'aws-sdk';
```
NGL321, line 1: While it contradicts the above comment, I think non-specific imports for the AWS SDK makes sense even if it is contradictory since the SDK is specifically designed for use with general imports.

typescript/api-cors-lambda-crud-dynamodb/lambdas/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../tsconfig.json"
+}
```
NGL321, line 3: New line needed

typescript/api-cors-lambda-crud-dynamodb/lambdas/update-one.ts
```diff
@@ -1,12 +1,14 @@
...
+
 const TABLE_NAME = process.env.TABLE_NAME || '';
 const PRIMARY_KEY = process.env.PRIMARY_KEY || '';
 
+
 const RESERVED_RESPONSE = `Error: You're using AWS reserved keywords as attributes`,
   DYNAMODB_EXECUTION_ERROR = `Error: Execution update, caused a Dynamodb error, please take a look at your CloudWatch Logs.`;
 
-export const handler = async (event: any = {}) : Promise <any> => {
+const db = new AWS.DynamoDB.DocumentClient();
```
NGL321, line 10: Imo should be separated by line break as was done in previous files.

typescript/api-cors-lambda-crud-dynamodb/tsconfig.json
```diff
@@ -15,7 +19,9 @@
...
-        "strictPropertyInitialization":false
-    }
-}
-
+        "strictPropertyInitialization": false
+    },
+    "include": [
+        "*.ts"
+    ]
+}
```
NGL321, line 27: Another newline

NGL321 (changes requested):
Looks really good. I have a couple notes for improvements, then its ready to merge

ryparker:
> Looks really good. I have a couple notes for improvements, then its ready to merge

Thanks for the suggestions. I implemented all of them and I went a bit further and cleaned up the index.ts file. Specifically the lambda props. Instead of duplicating them in each lambda declaration I stored them in `nodeJsFunctionProps` and reused them in each lambda declaration.

LMK what you think.

typescript/api-cors-lambda-crud-dynamodb/lambdas/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "license": "MIT",
+  "devDependencies": {
+    "@types/node": "*",
+    "@types/uuid": "*"
+  },
+  "dependencies": {
+    "aws-sdk": "*",
+    "uuid": "*"
+  }
+}
```
ryparker, line 15: Strange this should have a newline....

ryparker: For some reason I cannot see the newline in GitHub's review UI. However when I click "edit file" I see there is already a new line. LMK if you don't see it during your review @NGL321.

ryparker: Did some digging into this and it looks like every file does have a newline char added however you'll only see it in your local editor or when editing the file via GitHub's "edit file" UI. If you want the newline to render in GitHubs "view file" view, then you'll have to add 2 newlines to the bottom of the file.

Is this alright as is? @NGL321

NGL321 (approved):
Looks great!! 🚀

PR #487, opened 2021-08-12
feat(lambda-layer): Added python lambda layer cdk example

### LINKED ISSUE #486: Using Lambda Layers with Lambda Functions in CDK
### General Information

* [X] :wave: I may be able to implement this feature request

* [ ] :warning: This feature might incur a breaking change 

### Description

I'd like to add examples that demonstrate use of Lambda Layers with Lambda Functions as CDK components. In addition, demonstrate the language-specific folder structures that are used when create Lambda Layers as seen in [documentation](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path).
This use case is quite common for enterprise deployments but not included as an example here.

### Proposed Solution

I am happy to make a PR for multiple languages, starting with Python for initial review/baseline before others.

### Environment

  - **CDK CLI Version:** 1.117.0 (build 0047c98)
  - **Example:** lambda-layer (*new*)
  - **Example Version:** 
  - **OS:** all
  - **Language:** all

asadana:
Fixes #486 

---


alex9311:
hey @asadana I'm struggling to get this to work and found this example of yours. Do you know if the layer code has to be within a `python` directory?

My structure is:
```
python
---- lambda
-------- index.py
---- layer
-------- __init__.py
-------- helpers
------------ __init__.py
------------ connections.py
```
And my layer code asset is pointed to `layer`. Is it expected that importing from modules nested in layer dirs should work?

When I try to `from helpers.connection import my_func` I get `"Unable to import module 'index': No module named 'helpers'`. Perhaps the example could be more clear about what the required structure is?

asadana:
@alex9311 
While `layer/python` directory is not _required_ but it makes it much easier for automatic PYTHONPATH handling, everything inside the python directory can be referenced directly.
If you would like to reference a folder structure like helpers, you can place it under layer/python

---- lambda
-------- index.py
---- layer
-------- python
------------ helpers
---------------- __init__.py
---------------- connections.py

ryparker (approved):
Successfully deployed to AWS account.

PR #497, opened 2021-09-04
feat: new cdk python sample ec2-alarm-opsitem

alekasw:
### Description
This example shows how you can use the cdk to automate attaching an alarm to an ec2 instance at launch time, then when the alarm is triggered how it can create an systems manager opsitem with an associated run book.   

### Proposed Solution

Koi-Demo-Architecture

The workflow of this solution is as follows:  When an EC2 instance is launched it will trigger an Eventbridge rule that kicks off a lambda function.  The lambda function determines if the EC2 instance already has a matching alarm.  If it does **not** then it will create and attach a "StatusCheckFailed" metric alarm and tag the instance so next time it is launched it will skip the Alarm creation logic.  

The solution also deploys an SSM automation run command document that can be used to easily trigger the alarm via a bash 
shell script that executes the **set-alarm-state** aws cli command.  

Once the alarm is triggered another Eventbridge rule will kick-off the second lambda function that creates an SSM OpsItem with an associated runbook.  

**Clean-up:** cdk destroy then delete any Alarms that were created 

### Environment

  - **CDK CLI Version:** 1.121.0 
  - **Example:** Automate EC2 alarm creation that triggers an OpsItem when in "ALARM" state
  - **Example Version:** 1.0
  - **OS:** Amazon Linux 2
  - **Language:** Python

### Other information 
The cdk stack deploys the following resources:
 - IAM Policies and Roles
 - IAM Instance Profile
 - EC2 Instance
 - SNS Topic and Subscription
 - Lambda Functions (Two)
 - Eventbridge Rules (Two)
 - SSM Document


alekasw:
Not sure why the build phase is failing with "need to perform AWS calls for account but no credentials have been configured".  Can somebody advise?

indrora:
Thank you for your contribution. Unfortunately, this PR is very old, has had no movement, and predates the launch of CDK v2. 

I'm going to close this PR for now. If you update it to use the CDKv2, please open a new PR.

PR #557, opened 2021-11-02
feat(typescript): Simplify certificate logic

nerg4l:
In the current static page example for Typescript there is no need to create a custom `ICertificate` because `acm.DnsValidatedCertificate` already implements that interface. This PR removes the unnecessary extra step of creating an extra object and passes `acm.DnsValidatedCertificate` directly to `cloudfront.ViewerCertificate.fromAcmCertificate` as the first parameter.

---


ryparker:
Looks like there's a merge conflict here. Would you be able to resolve this?

Thanks for contributing 🙏🏻

ryparker (approved):

PR #624, opened 2022-03-10
feat (Typescript): Update static site example to use new Distribution API

### LINKED ISSUE #609: (aws-cdk-examples): please update the static-site example to use the modern cloudfront Distribution API
### General Information

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#cloudfrontwebdistribution-api states that "The CloudFrontWebDistribution construct is the original construct written for working with CloudFront distributions. Users are encouraged to use the newer Distribution instead, as it has a simpler interface and receives new features faster." However, the example here uses the older API. I was looking for an example because I'm getting a lot of errors connecting CloudFront to an S3 bucket, but the mix of the older API here with the newer documentation doesn't help.

### Proposed Solution

Update the typescript/static-site example

### Language

Typescript

### Other information

_No response_

### Acknowledge

- [ ] I may be able to implement this feature request

its-mirus-lu:
This PR updates the static-sites example to use the new Distribution API following the migration instructions outlined here: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#migrating-from-the-original-cloudfrontwebdistribution-to-the-newer-distribution-construct.

The changes replace the CloudFrontWebDistribution with Distribution and modifies the way that the certificate is associated with the Distribution.

Fixes #609 

---


its-mirus-lu:
This is a messy PR, pulling in work that is outside of the intended merge. Will re-raise a new PR shortly

PR #625, opened 2022-03-10
feat(Typescript): Updated static-site example to use new Distribution API

its-mirus-lu:
This PR updates the static-sites example to use the new Distribution API following the migration instructions outlined here: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#migrating-from-the-original-cloudfrontwebdistribution-to-the-newer-distribution-construct.

The changes replace the CloudFrontWebDistribution with Distribution and modifies the way that the certificate is associated with the Distribution.

Fixes https://github.com/aws-samples/aws-cdk-examples/issues/609

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #632, opened 2022-03-23
fix: Addressed issue where distribution ViewerCertificate attributes were being placed in wrong place in Distribution hierarchy

### LINKED ISSUE #609: (aws-cdk-examples): please update the static-site example to use the modern cloudfront Distribution API

its-mirus-lu:
*Requesting @peterwoodworth review this please*

This PR addresses an issue that was introduced in PR #625... the issue is related to the placement of the ViewerCertificate attributes in a Distribution resource.  From below we see the original synthesized CFN
```
"StaticSiteSiteDistributionE3AF6299": {
      "Type": "AWS::CloudFront::Distribution",
      "Properties": {
        "DistributionConfig": {
          "Aliases": [
            ...
          ],
          "DefaultCacheBehavior": {
            ...
          "Origins": [
            ...
          ],
          "ViewerCertificate": {
            "AcmCertificateArn": {
              "Fn::GetAtt": [
                "StaticSiteSiteCertificateCertificateRequestorResourceB92D481B",
                "Arn"
              ]
            },
            "MinimumProtocolVersion": "TLSv1.2_2021",
            "SslSupportMethod": "sni-only"
          }
        },
        "ViewerCertificate": {
          "MinimumProtocolVersion": "TLS_V1_1_2016",
          "SslSupportMethod": "sni-only"
        }
      },
    },
```
From the above, which was generated from the code in my previous commit, the MinimumProtocolVersion and SslSupportMethod attributes were being placed in the high-level Distribution resource when it should be placed within the DistributionConfig property of the resource.

Going into static-site.ts, from the code block where we use addPropertyOverride, I did this:

cfnDistribution.addPropertyOverride('MinimumProtocolVersion', 'TLS_V1_1_2016') cfnDistribution.addPropertyOverride('ViewerCertificate.SslSupportMethod', 'sni-only')

when it should have been this:

cfnDistribution.addPropertyOverride('DistributionConfig.ViewerCertificate.MinimumProtocolVersion', 'TLSv1.1_2016') cfnDistribution.addPropertyOverride('DistributionConfig.ViewerCertificate.SslSupportMethod', 'sni-only')

I bought a $3 domain in R53 and validated deployment in my own AWS account.

Fixes #609

---

### DIFF
--- a/typescript/static-site/static-site.ts
+++ b/typescript/static-site/static-site.ts
@@ -6,7 +6,7 @@ import * as cloudfront from 'aws-cdk-lib/aws-cloudfront';
 import * as s3deploy from 'aws-cdk-lib/aws-s3-deployment';
 import * as targets from 'aws-cdk-lib/aws-route53-targets';
 import * as cloudfront_origins from 'aws-cdk-lib/aws-cloudfront-origins';
-import { CfnOutput, RemovalPolicy, Stack } from 'aws-cdk-lib';
+import { CfnOutput, Duration, RemovalPolicy, Stack } from 'aws-cdk-lib';
 import * as iam from 'aws-cdk-lib/aws-iam';
 import { Construct } from 'constructs';
 
@@ -36,8 +36,6 @@ export class StaticSite extends Construct {
     // Content bucket
     const siteBucket = new s3.Bucket(this, 'SiteBucket', {
       bucketName: siteDomain,
-      // websiteIndexDocument: 'index.html',
-      // websiteErrorDocument: 'error.html',
       publicReadAccess: false,
       blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
 
@@ -78,6 +76,14 @@ export class StaticSite extends Construct {
       defaultRootObject: "index.html",
       domainNames: [siteDomain],
       minimumProtocolVersion: cloudfront.SecurityPolicyProtocol.TLS_V1_2_2021,
+      errorResponses:[
+        {
+          httpStatus: 403,
+          responseHttpStatus: 403,
+          responsePagePath: '/error.html',
+          ttl: Duration.minutes(30),
+        }
+      ],
       defaultBehavior: {
         origin: new cloudfront_origins.S3Origin(siteBucket, {originAccessIdentity: cloudfrontOAI}),
         compress: true,
@@ -85,8 +91,6 @@ export class StaticSite extends Construct {
         viewerProtocolPolicy: cloudfront.ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
       }
     })
-
-    
 
     new CfnOutput(this, 'DistributionId', { value: distribution.distributionId });
 
--- a/typescript/static-site/static-site.ts
+++ b/typescript/static-site/static-site.ts
@@ -86,12 +86,7 @@ export class StaticSite extends Construct {
       }
     })
 
-    // This is a workaround for https://github.com/aws/aws-cdk/issues/19539
-    // const cfnDistribution = distribution.node.defaultChild as cloudfront.CfnDistribution
-    // cfnDistribution.addDeletionOverride('DistributionConfig.Origins.0.CustomOriginConfig')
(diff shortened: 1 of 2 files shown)


peterwoodworth (dismissed):

typescript/static-site/static-site.ts
```diff
@@ -82,8 +82,8 @@ export class StaticSite extends Construct {
     })
 
     const cfnDistribution = distribution.node.defaultChild as cloudfront.CfnDistribution
-    cfnDistribution.addPropertyOverride('MinimumProtocolVersion', 'TLS_V1_1_2016')
-    cfnDistribution.addPropertyOverride('ViewerCertificate.SslSupportMethod', 'sni-only')
+    cfnDistribution.addPropertyOverride('DistributionConfig.ViewerCertificate.MinimumProtocolVersion', 'TLSv1.1_2016')
+    cfnDistribution.addPropertyOverride('DistributionConfig.ViewerCertificate.SslSupportMethod', 'sni-only')
```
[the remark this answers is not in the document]
its-mirus-lu: You're completely correct. I was doing a 1:1 port of the v1 constructs to the v2 constructs.  The previous version was explicitly setting TLSv1.1_2016.  I'll remove the reference which will implicitly set the version to TLSv1.2_2021.

its-mirus-lu:
@peterwoodworth in passing, I've been trying to connect to the website that was deployed and I'm getting a 403 error.  It looks like a couple of things are a bit off: 

1) CDK is synthing a cfn template with an origin that doesn't have an OAI attached
2) CDK is synthing a domain for the s3 bucket that is the website url instead of the REST url (which is required if public access is disabled)

I'm trying to find a workaround and raise a ticket to the CDK project itself.

I know you must be busy, but when you have a free moment, can you try to deploy a s3 hosted static website + cloudfront using CDK v2? I need a second opinion to make sure I'm not crazy.

its-mirus-lu:
Added errorResponses and also changed the backing bucket to have website hosting disabled -- this would allow the S3Origin to be configured properly -- see this [bug](https://github.com/aws/aws-cdk/issues/19539#) for details.

peterwoodworth:
Thanks for the thorough bug report - makes sense. Really appreciate your effort here @its-mirus-lu

typescript/static-site/static-site.ts
```diff
@@ -70,21 +69,29 @@ export class StaticSite extends Construct {
     });
     new CfnOutput(this, 'Certificate', { value: certificate.certificateArn });
 
+    
     // CloudFront distribution
     const distribution = new cloudfront.Distribution(this, 'SiteDistribution', {
       certificate: certificate,
+      defaultRootObject: "index.html",
       domainNames: [siteDomain],
+      minimumProtocolVersion: cloudfront.SecurityPolicyProtocol.TLS_V1_2_2021,
```
peterwoodworth, line 78: This line isn't completely necessary - this value will get set by default if you don't include this prop, but it doesn't hurt I guess.

peterwoodworth (approved):

typescript/static-site/static-site.ts
its-mirus-lu: Good point. I'm noticing that it also gets set as a feature flag (for lack of a better term) in cdk.json.  I guess leaving it in would make it more explicit for future readers of this code.

PR #662, opened 2022-04-19
feat(typescript): Adds CDK example for setting up S3 cross account replication with KMS encrypted buckets

### LINKED ISSUE #661: Use CDK to setup Cross Account S3 Replication with KMS Encryption
### Overview
A useful feature that I've seen in the past is to replicate S3 objects from one bucket to another bucket. 
AWS also recommends encrypting your S3 buckets as standard practice. 
I found several examples of replicating objects, or even replicating encrypted objects but none of how to perform cross account replication with KMS encrypted buckets. 

### Justification
The reason I believe this is a good example is that it illustrates the following concepts:
S3 Object Replication
KMS Encrypted S3 Buckets
S3 Cross Account Replication while maintaining encrypted objects and transferring ownership of the new objects to the destination bucket.

### Usecase
1. System in account A creates a report file in an S3 KMS encrypted bucket (the source).
1. System in account B needs to consume the report created in account A in step 1. 
1. Account A S3 source bucket is configured for S3 replication to replicate to account B S3 destination bucket.
1. Replication Role for account A source bucket needs to have permission to read objects and decrypt them in account A using the KMS encryption key. 
1. Replication Role for account A needs to have permission to write, replicate and reencrypt the objects using the KMS key in account B. 
1. Replication Role for account A needs to have permission to transfer the ownership of the S3 object to account B. 

### Sequence
1. Account A System/User uploads file to S3 bucket in account A. 
1. S3 replicates file from account A to account B while maintaining encryption and transferring ownership of the object to account B. 
1. Account B System/User reads file from S3 bucket in account B.

### Proposed Solution

Setup S3 Replication Rules to replicate the S3 objects across the account boundary using S3 while decrypting at the source and automatically re-encrypting at the destination.

### Other Information

_No response_

### Acknowledgements

### Language

Typescript

mastephens:
Fixes #661 

---


typescript/s3-kms-cross-account-replication/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@types/jest": "^26.0.10",
+    "@types/node": "10.17.27",
+    "aws-cdk": "2.19.0",
+    "jest": "^26.4.2",
+    "ts-jest": "^26.2.0",
+    "ts-node": "^9.0.0",
+    "typescript": "~3.9.7"
+  },
+  "dependencies": {
+    "@aws-cdk/aws-iam": "^1.151.0",
```
rayjanwilson, line 23: since you pulled in `aws-cdk-lib`,you don't need to pull the `@aws-cdk/***` libraries

typescript/s3-kms-cross-account-replication/stacks/step2-destination-account.ts
```diff
@@ -0,0 +1,137 @@
...
+      encryption: BucketEncryption.KMS,
+      encryptionKey: destinationKmsKey
+    });
+
+    // allow the principal to have all admin access to bucket
+    destinationS3Bucket.addToResourcePolicy(
+      new PolicyStatement({
+        sid: 'Set Admin Access',
+        effect: Effect.ALLOW,
+        principals: [new ArnPrincipal("arn:aws:iam::" + Config.destinationAccountId + ":root")],
```
rayjanwilson, line 69: consider backticking this rather than using the overloaded addition: 
`arn:aws:iam::${Config.destinationAccountId}:root`

typescript/s3-kms-cross-account-replication/package.json
mastephens: Thanks! I removed the 1.x dependencies

typescript/s3-kms-cross-account-replication/stacks/step2-destination-account.ts
mastephens: Thanks, updated across file for consistency

ryansonshine:
@NGL321 can we get a review on this when you get a chance?

Thanks! :smile:

rayjanwilson:
looking for a response to when we might be able to get a review on this

peterwoodworth:
Hey all, thanks for your patience, and for the contribution! We'll try to take a look at this in the next week

mastephens:
Are there any outstanding questions on this PR? Thanks!

peterwoodworth:
Yeah, I'll have some comments about this. Will finish the review tomorrow 🙂

peterwoodworth (commented):
I'm going to get to a couple finer details tomorrow, but I'll post something today since I said I would 🙂 

Overall, this is excellent! Easy to deploy, and _great_ work on implementing the cross region replication through policies and escape hatches. (Sidenote: If you have any opinions on how the CDK can implement this feature, please leave a comment on [this issue](https://github.com/aws/aws-cdk/issues/1680))

I have two concerns over how this is configured and deployed

First thing, as a best practice, CDK recommends against specifying the physical names of objects when possible. 
This is largely because if you make any modifications to a resource which requires resource replacement *and* you've specified its name, you will reach a [pretty undesirable state](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html) 
I would prefer to leave these options in the config file both optional, and undefined by default.

Second, since all these stacks are working within the same app, we can pass the values between the stacks instead of relying on CloudFormation parameters. This makes the deployment step a bit easier and harder to mess up! What do you think about this?

mastephens:
@peterwoodworth Thanks for taking the time to do a review, this feedback is excellent.

1: Static Names
I agree, it would be preferred to have dynamic names generated. 
In this scenario there is a circular reference occurring between the replication role and the buckets/keys.
Because the role needs a policy that contains the bucket names in order to replicate, the stack needs a known name to define for the policy. 
The buckets, and in turn the keys need to be known for the buckets to set in their own policies, and the keys need the replication role defined in the first step for their own policies.
If there is a better way to accomplish this circular reference using a CDK function I would appreciate any insight you might be able to point me to.

I could easily remove the values from the config files by default.
A side effect will be that the unit tests need the values to succeed and I'll need an alternate configuration available for them to succeed.

2: Cross Account Stack References
I had previous tried to get this functioning as a single stack that can be deployed as one unified app and referencing either the variables defined on the previous stack or trying to import the CfnOutputs into the next stack. 
In both cases there were issues with moving the variables across account boundaries that prevented me from succeeding. 

In this case I thought that perhaps it was better to have a functioning demonstration on how to setup x-account replication of encrypted buckets rather than the more complex demonstration of CDK x-account resource sharing.
I will make additional efforts today to get this working to outline the specific issues that occurred.
If you have an example of how this could be done that I could review I would appreciate it.

Thanks!

peterwoodworth:
Hey, gonna get back to this in a few hours 🙂 Thanks again for your patience!

peterwoodworth (approved):
I've thought about it and tested it a bit - and you're right about both of the concerns you have. With the cross account nature of this, this should involve parameters anyway.

Also, you're right about how we need to have the names defined due to the circular reference.

This PR is excellent. I really appreciate the patience and the care that went into this PR before we had a chance to look at it. Thanks a bunch 🙂

PR #665, opened 2022-04-26
feat(python): adding cdk example serverless-backend api gateway, lamb…

### LINKED ISSUE #664: serverless-backend
### Describe the feature

Setting up a serverless backend with: 

- Amazon API Gateway
- AWS Lambda
- Amazon S3
- Amazon DynamoDB
- Amazon Cognito
- AWS IAM

### Use Case

User makes an authenticated API call to POST an image/file and its metadata in a JSON request. Lambda receives the request and processes the image, and stores it in an S3 bucket, while its metadata is processed and stored in a DynamoDB table. 

### Proposed Solution

I am proposing to build a CDK sample in Python that illustrates how to build a serverless backend application that will allow users to make authenticated API calls with API Gateway and Cognito and have the request be processed by a Lambda. Will use S3 to store images/files and DynamoDB to store any metadata needed. 

### Other Information

_No response_

### Acknowledgements

### Language

Python

Amralsa:
Fixes #664 

---


python/serverless-backend/README.md
```diff
@@ -0,0 +1,78 @@
...
+                The password for the new user needs to be permanent. To do so, run the following command (Change the value of user-pool-id to the deployed Cognito User Pool ID. Change the username and password to your created values):
+                aws cognito-idp admin-set-user-password \
+                --user-pool-id us-east-1_xxxxxx\
+                --username usernameExample \
+                --password passwordExample \
+                --permanent
+                Next, the IdToken of the account needs to be retrieved. To do so, run the following command (Change the value of client-id to the deployed Cognito User Pool Client ID. Change the username and password to your created values):
+                aws cognito-idp initiate-auth —region us-east-1 —auth-flow USER_PASSWORD_AUTH —client-id XXXXXXXXXXXXXXXXXX —auth-parameters USERNAME=usernameExample,PASSWORD=passwordExample
+                This command will return multiple keys. Save the IdToken for later use
+            - Use Postman to test your API endpoint
```
indrora, line 53: Nitpick: a cURL call here would potentially be more useful.

python/serverless-backend/README.md
```diff
@@ -0,0 +1,78 @@
...
+            7. Go to your terminal to run some AWS CLI commands. 
+
+                The password for the new user needs to be permanent. To do so, run the following command (Change the value of user-pool-id to the deployed Cognito User Pool ID. Change the username and password to your created values):
+                aws cognito-idp admin-set-user-password \
+                --user-pool-id us-east-1_xxxxxx\
+                --username usernameExample \
+                --password passwordExample \
+                --permanent
+                Next, the IdToken of the account needs to be retrieved. To do so, run the following command (Change the value of client-id to the deployed Cognito User Pool Client ID. Change the username and password to your created values):
+                aws cognito-idp initiate-auth —region us-east-1 —auth-flow USER_PASSWORD_AUTH —client-id XXXXXXXXXXXXXXXXXX —auth-parameters USERNAME=usernameExample,PASSWORD=passwordExample
```
indrora, line 51: Make it clear that you need to use the region that the stack was deployed in

python/serverless-backend/README.md
```diff
@@ -0,0 +1,78 @@
+# Serverless-backend for your UI / Frontend example
+
+This is a project to deploy a serverless backend for your current project on AWS with the AWS Cloud Development Kit.
+
+
+
+This project also demonstrates how to build a serverless backend application that will allow users to make authenticated API calls with API Gateway and Cognito and have the request be processed by a Lambda. Will use S3 to store images/files and DynamoDB to store any metadata needed.
```
indrora, line 7: Clarify and shorten this: Tell me what I'll be able to do when I've deployed this stack & what is happening behind the scenes.

python/serverless-backend/source.bat
```diff
@@ -0,0 +1,13 @@
+@echo off
```
indrora, line 1: Don't include venv stuff.

indrora (changes requested):
Looks good overall but needs documentation and some cleanup. 

For an example that demonstrates best practices, see https://github.com/aws-samples/aws-cdk-examples/tree/master/python/image-content-search

indrora:
Nearly there, just a few small things to bring it in line with our typical example.

python/serverless-backend/README.md
Amralsa: Using Postman here because it's easier to send image in a base64 format.

Amralsa (commented):
resolved issues.

indrora:
A few things now that I've had some caffeine in me... Coffee makes the brain go good.

This sample should include a warning about file size:

- Lambda has a maximum [6MB payload limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html), which is smaller than API Gateway's 10MB limit.
- Because the content has been base64 encoded, it suffers a [roughly 33% bloat in size](https://developer.mozilla.org/en-US/docs/Glossary/Base64), limiting the effective size of any file uploaded through this to be about 4MB -- more than likely 1-2MB in size for a 90% JPEG with high noise. This is *before* compression.
- Customers that want >2-4MB should know about this limitation.

The correct pattern for this would be:

* Sending the initial lambda-API Gateway request and being [returned a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html)
* The frontend uploading a file via the presigned URL (using PUT)
* S3 invoking a lambda function to mark that the file has been uploaded successfully (setting, for instance, a mimetype in DynamoDB)

Amralsa:
> A few things now that I've had some caffeine in me... Coffee makes the brain go good.
> 
> This sample should include a warning about file size:
> 
> * Lambda has a maximum [6MB payload limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html), which is smaller than API Gateway's 10MB limit.
> * Because the content has been base64 encoded, it suffers a [roughly 33% bloat in size](https://developer.mozilla.org/en-US/docs/Glossary/Base64), limiting the effective size of any file uploaded through this to be about 4MB -- more than likely 1-2MB in size for a 90% JPEG with high noise. This is _before_ compression.
> * Customers that want >2-4MB should know about this limitation.
> 
> The correct pattern for this would be:
> 
> * Sending the initial lambda-API Gateway request and being [returned a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html)
> * The frontend uploading a file via the presigned URL (using PUT)
> * S3 invoking a lambda function to mark that the file has been uploaded successfully (setting, for instance, a mimetype in DynamoDB)

Agreed, I will add a warning and outline the limitations of this solution under prerequisites.  The pre-signed URL solution is another way to upload an image or a file, undoubtedly. However, it introduces complexity and control limitations.  1) The user has the freedom to upload any size of an object that is up to 5GB with no restrictions. 2) Within the expiration period, the presigned URL can be used by any user, regardless of who they are to upload an object, which introduces a security risk. There are several parameters that cannot be embedded within the URL, using the `generate_presigned_url` for example. 


A third solution, which is similar to your purpose one in which it uses multiple API calls to make the POST request, is the Presigned POST solution. Here, we can directly call the S3 API from the client side,  restrict the size of the object, add a POST policy, and more. However, this also introduces complexity such as CORS, IAM permissions, etc., for a CDK example that is looking to get the user started with using CDK to build and integrate their infrastructure. 


The current solution is the most simpler/straightforward solution of uploading an image/file with a maximum size of 2MB. As you have mentioned above, it does come with restrictions and limitations since Lambda is proxying the PUT action here. The customer can use this example, to see how we can: 1) have a file/image storage that is integrated with Lambda. 2) have a metadata storage to store any type of data in key/value form.

indrora (approved):
Addition of limitations to the README satisfies my needs. 👍 LGTM

indrora (approved):

indrora (approved):

PR #674, opened 2022-05-12
fix(python) Remove unsafe python example

indrora:
This removes the unsafe lambda-ddb-mysql-etl-pipeline example, which used inline SQL generation and did not follow best practices w/r/t relational DB access. 

This will be rewritten in the future.


NGL321 (approved):
Seems clear-cut with good reasoning

PR #676, opened 2022-05-13
chore(deps): update typescript to 4.6.0

corymhall:
re https://github.com/aws/aws-cdk/issues/20319

The other option is to do what is mentioned in the referenced issue and
add a devDependency for `@types/prettier`, but since it is possible to
upgrade `typescript` that seems like the better option.

Fixes # 

---


Chriscbr (approved):
LGTM

PR #689, opened 2022-06-16
Allow cdk.context.json to be committed

wgiddens:
The CDK docs state cdk.context.json file should be committed. Otherwise you may run into inconsistent deployments.

See https://docs.aws.amazon.com/cdk/v2/guide/context.html#:~:text=Because%20they%20are%20part%20of%20your%20application%27s%20state%2C%20cdk.json%20and%20cdk.context.json

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -375,7 +375,6 @@ $RECYCLE.BIN/
 
 ### CDK-specific ignores ###
 *.swp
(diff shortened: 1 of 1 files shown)


indrora:
cdk.context.json contains information that is specific to your account configuration and is not committed in this repository.

PR #691, opened 2022-06-19
Update s3_sns_sqs_lambda_chain_stack.py

pmithrandir:
The visibility timeout is set to 30 seconds, which is fine... only if the lambda timeout is 5 sec.

According to that doc, there should be a factor 6 between the 2 values.
https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

Not doing so can result in a "loop" in the lambda activation by SQS, leading to large overbill in no time.

This rule is quite hrd to find in AWS documentation, so I think it's better to enhance the example in order to underline it.

Pierre

Fixes # 

---


indrora (dismissed):

indrora (approved):

python/s3-sns-sqs-lambda-chain/s3_sns_sqs_lambda_chain/s3_sns_sqs_lambda_chain_stack.py
```diff
@@ -35,7 +36,8 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
       self,
       id="sample_queue_id",
       visibility_timeout=Duration.seconds(30),
```
mmaeng, line 38: Looks like you have defined the same parameter twice.

PR #692, opened 2022-06-29
chore: fix broken build in RDS Aurora example

rix0rrr:
The code was bypassing null-checks, which of course led to
passing an `undefined` value which breaks the CDK synthesis.

---

### DIFF
--- a/typescript/rds/aurora/aurora.ts
+++ b/typescript/rds/aurora/aurora.ts
@@ -142,7 +142,7 @@ export class Aurora extends Stack {
     let subnetIds = props.subnetIds;
     let instanceType = props.instanceType;
     let replicaInstances = props.replicaInstances;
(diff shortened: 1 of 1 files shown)


typescript/rds/aurora/aurora.ts
```diff
@@ -154,7 +154,7 @@ export class Aurora extends Stack {
       throw new Error('Unknown Engine Please Use mysql or postgresql');
       process.exit(1);
     }
-    if (backupRetentionDays! < 14) {
+    if (backupRetentionDays < 14) {
       backupRetentionDays = 14;
     }
     if (replicaInstances! < 1) {
```
otaviomacedo, line 160: Should we do the same here?

rix0rrr: Good call

RomainMuller (approved):

Issue #698, opened 2022-07-07
Missing dependencies in appsync-graphql-dynamodb example


samputer:
### Describe the bug

Cannot tear down `appsync-graphql-dynamodb` due to missing dependencies on datasource.

### Expected Behavior

Should be able to destroy stack when running `cdk destroy`

### Current Behavior

`cdk destroy` yields the following error

```
16:24:39 | DELETE_FAILED        | AWS::AppSync::DataSource    | ItemsDataSource
Resolver for type Mutation and field delete associated with data source (Service: AWSAppSync; Status Code: 400; Error Code: BadRequestException; Request ID: ...
```

### Reproduction Steps

In the `typescript/appsync-graphql-dynamodb` directory:

`cdk deploy`

followed by

`cdk destroy`

Observe error

### Possible Solution

Add missing dependencies to all of the datasources. PR inbound.

### CDK CLI Version

2.31.0 (build b67950d)

### Node.js Version

v16.13.1

### OS

MacOS 12.4

### Language

Typescript

samputer:
PR raised: https://github.com/aws-samples/aws-cdk-examples/pull/699

awscamille:
I think #1005 addresses that issue.

kaizio:
Confirmed one is able to `cdk deploy` and `cdk destroy`.

PR #699, opened 2022-07-07
Add missing dependencies to appsync-graphql-dynamodb example

samputer:
Fixes inability to destroy appsync-graphql-dynamodb example

Fixes #
https://github.com/aws-samples/aws-cdk-examples/issues/698
---

### DIFF
--- a/typescript/appsync-graphql-dynamodb/index.ts
+++ b/typescript/appsync-graphql-dynamodb/index.ts
@@ -91,6 +91,7 @@ export class AppSyncCdkStack extends cdk.Stack {
       responseMappingTemplate: `$util.toJson($ctx.result)`
     });
     getOneResolver.addDependsOn(apiSchema);
+    getOneResolver.addDependsOn(dataSource);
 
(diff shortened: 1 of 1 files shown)


peterwoodworth (commented):
I actually wasn't able to deploy due to this either, and before I looked at the PR I made the exact changes you made so that I could deploy to test what would happen on destroy 😆 

That said, while this fix works, I think this example overall has fallen below our standards. We shouldn't have to use the `Cfn` resources anymore because we have a modern [Appsync library](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-appsync-readme.html) which if used, would make the code here more intuitive and stable. I would like to migrate this example to use our Appsync module rather than directly using CloudFormation's Appsync resources. Would you be able to migrate this example over? No worries if not, just let me know

peterwoodworth (changes requested):
Regardless of your answer to my previous question, I'd like to see if you can implement this fix in all of the languages where this example exists 🙂

Issue #32, opened 2019-04-16
Figure out why override with null is causing an error


rix0rrr:
See

https://github.com/aws-samples/aws-cdk-examples/blob/master/typescript/resource-overrides/index.ts#L55

PR #738, opened 2022-09-20
feat: Add csharp/resource-overrides 

kaizio:
Expanded on and got working #175, which adds Csharp port of the TS CDK app resource-overrides. This also updates the TS version to the latest version of CDK.

Fixes #175

---


kaizio:
Typescript update to v2 that this includes Fixes #32. 

No longer get undefined for the filter. If you enable the line you now get CDK reporting a error. I have removed the line since no longer needed.

```
 ❌  resource-overrides failed: Error [ValidationError]: [/Resources/MyBucketF68F3FF0/Type/Foo] 'null' values are not 
  allowed in templates
```

PR #760, opened 2022-10-07
fix: Java snapshot test is too brittle

rix0rrr:
This test broke because a recent release added a dependency. 

Just check the important part of the template using the `assertions` library.

---


Naumel (approved):
🪴

Naumel (approved):

PR #783, opened 2022-12-15
feat(typescript): Add new CDK example demonstrating how to use EC2 instances in a Local Zone with AWS Systems Manager

markusl:
Add new CDK example demonstrating how to use EC2 instances in a Local Zone with AWS Systems Manager

---


pahud:
Hi @markusl 

Thanks for your contribution. I'll look into this and try deploy it in my account.

typescript/ec2-ssm-local-zone/example-localzone-stack.ts
```diff
@@ -0,0 +1,204 @@
...
+
+// TODO: Change these to suitable values
+const PARENT_REGION_AZ = 'eu-north-1a';
+const LOCAL_ZONE_AZ = 'eu-north-1-hel-1a';
+
+const VPC_CIDR = '172.31.100.0/24';
+const SUBNET_SIZE = 26;
+
+// Amazon Linux 2022 https://docs.aws.amazon.com/linux/al2022/ug/what-is-amazon-linux.html
+const al2022Ami = ec2.MachineImage.fromSsmParameter('/aws/service/ami-amazon-linux-latest/al2022-ami-kernel-5.15-x86_64');
```
pahud, line 16: Any reason not using this instead?

```ts
const al2022Ami = new ec2.AmazonLinuxImage({
  generation: ec2.AmazonLinuxGeneration.AMAZON_LINUX_2022,
})
```

pahud (commented):
Thank you for your PR. It deploys in my account.

Is there any reason not using ALB and AutoScaling group in local zone? Understand the NAT gateway is not currently supported in local zone but having two ASG(one for NAT instance and one for nginx instance) seems to be a good idea for high availability concern. Thoughts?

typescript/ec2-ssm-local-zone/example-localzone-stack.ts
pahud: OK there seems to be a bug in `AMAZON_LINUX_2022` that always returns kernel 5.10 whereas only 5.15 is currently available. 

https://github.com/aws/aws-cdk/blob/cea1039e3664fdfa89c6f00cdaeb1a0185a12678/packages/%40aws-cdk/aws-ec2/lib/machine-image.ts#L392

https://github.com/aws/aws-cdk/blob/cea1039e3664fdfa89c6f00cdaeb1a0185a12678/packages/%40aws-cdk/aws-ec2/lib/machine-image.ts#L467

I'll see what we can do as a workaround before we fix it from the upstream.

markusl: Updated!

markusl:
@pahud ALB is currently only supported in the US Local Zones. I added a note about this in the README

pahud:
LGTM. I am leaving @peterwoodworth for another look and approval. Thank you for this PR.

markusl:
Build fails at ```
lib/s3-object-lambda-stack.ts:92:15 - error TS2322: Type '{ AwsLambda: { FunctionArn: string; }; }' is not assignable to type 'IResolvable | ContentTransformationProperty'.
  Object literal may only specify known properties, and ''AwsLambda'' does not exist in type 'IResolvable | ContentTransformationProperty'.``` which is a CDK v1 project which I think should be fixed separately.

peterwoodworth (approved):
Looks great, thanks Pahud for the initial review and @markusl for your hard work and patience 🙂

Issue #784, opened 2022-12-19
CDK Example: How to grant least privilege permission to a Secrets from a sample lambda 


agairola:
### Describe the feature

Writing KMS key policies can be complex because they can contain multiple statements that specify different permissions for different users and services. This can make it challenging for developer to understand the overall permissions granted by the policy, especially if the policy is long or contains many statements. Overall, creating a key policy for KMS requires a thorough understanding of the policy syntax, the different permissions and actions that can be specified, and how to effectively combine these elements to create a policy that meets the needs of your organization.

To solve this, a common pattern that I have used to create a effective way to grant least privilege permission to a sample lambda execution role using [grantRead](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-secretsmanager.Secret.html#grantwbrreadgrantee-versionstages) CDK method and [kms:ViaService](https://docs.aws.amazon.com/kms/latest/developerguide/policy-conditions.html#conditions-kms-via-service) condition. 

### Use Case

Developers are not always happy when they have to write least privilege IAM or KMS policies because it can be a time-consuming and tedious process. These policies are designed to limit access to resources and privileges within an organization's AWS account, which is important for security and compliance purposes. However, implementing these policies often requires a thorough understanding of the specific permissions and resources that are needed for an application or service to function properly, as well as a clear understanding of the organization's security and compliance requirements. This can be challenging for developers, particularly if they are not familiar with the organization's security and compliance policies or if they are working on a project with complex permissions requirements. Additionally, writing least privilege policies may require developers to make trade-offs between security and convenience, which can be frustrating and may require additional time and effort to get right.

### Proposed Solution

I am proposing to build a CDK sample in Typescript that illustrates how to build an efficient way of implementing the least privilege KMS policy for an AWS services. I have the code for this solution written and tested and will fork the repository to PR my solution.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Language

Typescript

PR #785, opened 2022-12-19
feat(typescript): Adding CDK example to showcase least privilege for KMS and ASM

agairola:
Fixes https://github.com/aws-samples/aws-cdk-examples/issues/784

---


typescript/kms-least-priv-asm/lambda/lambda_function.py
```diff
@@ -0,0 +1,61 @@
+import logging
```
markusl, line 1: Wouldn't it make sense to implement TypeScript example in TypeScript ;)

PR #786, opened 2022-12-21
feat(python): EventBridge-Lambda Construct and Service abstraction boiler plate

NoahLiot:
This is a construct and attached example which create a dummy docker image Lambda function and associate it with an [Amazon EventBridge](https://aws.amazon.com/eventbridge/) bus rule based on the service name in an existing EventBridge bus passed as a parameter of the construct. Additionally, the construct create a unique Dead Letter Queue using SQS.

This construct can be used as a fundamental building block to build, for example, a serverless microservices choregraphy. This is possible because by using Amazon EventBridge, we are able to leverage [Lambda asynchronous invocation](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html). Outside of direct the benefits of asynchronous invocation, this allows to chains services using the `on_success` and `on_failure` destination [documentation](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-destinations) with a provided Dead Letter Queue SQS in the construct for the `on_failure`. This pattern is currently the only way to fully respect the Infrastructure as Code paradigm. Indeed, in this configuration, and as per this construct, the service implementation does not need to be stateful regarding the input and output endpoints. Those outputs are defined directly in the CDK script while the microservices business logic can be written, run, tested, and deployed independently using containerization.

In this complete example, we are creating two services in the way described above, one which write back to its original source bus, and another which write a different bus. Note that the construct uses the `id` of the services to find the right path to the Docker file to build.

---


kaizio:
@NoahLiot Can you update your PR branch with the latest? So that we can verify with our new build processes.

PR #792, opened 2023-01-17
feat(typescript): Evidently Client-side evaluation for ECS

colethienes:
This is a CDK example for deploying AWS Evidently Client-side evaluation after the recent feature release: https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-integration-containers-agent.html

Deployed this into an AWS account and confirmed it works as expected.


indrora (approved):
Bonus: provide a screenshot/more context on where "oldSearchBar" will be.

PR #807, opened 2023-03-02
issue-615

michaelbwebb:
(typescript): An example using EventBridge rule to schedule Lambda function

Fixes # 

---


typescript/eventbridge-lambda/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import cdk = require('aws-cdk-lib');
+
+import fs = require('fs');
+
+export class EventBridgeLambdaStack extends cdk.Stack {
+  constructor(app: cdk.App, id: string) {
+    super(app, id);
+
+    const lambdaFn = new lambda.Function(this, 'Singleton', {
+      code: new lambda.InlineCode(fs.readFileSync('lambda-handler.py', { encoding: 'utf-8' })),
```
indrora, line 13: nitpick: `code.FromAsset` is generally a more "preferred" path, but this works.

typescript/eventbridge-lambda/lambda-handler.py
```diff
@@ -0,0 +1,2 @@
+def main(event, context):
+    print("I'm running!")
```
indrora, line 2: It would be really nice if this did something Useful:tm:

typescript/eventbridge-lambda/README.md
```diff
@@ -0,0 +1,35 @@
...
+npm install
+npm run build
+```
+
+This will install the necessary CDK, then this example's dependencies, and then build your TypeScript files and your CloudFormation template.
+
+## Deploy
+
+Run `cdk deploy`. This will deploy / redeploy your Stack to your AWS Account.
+
```
indrora, line 32: And how do you know it worked? How does one see that the lambda function is being called?

indrora (changes requested):

typescript/eventbridge-lambda/lambda-handler.py
michaelbwebb: I have updated this to run a lambda that kicks off an sns topic and sends an email to the user who initiated the CDK stack

typescript/eventbridge-lambda/README.md
michaelbwebb: Addressed in latest commit with sns topic subscription so you can get confirmation that the stack is working

michaelbwebb:
@indrora I hope these new updates are what you are looking for. Let me know if I am missing anything

michaelbwebb:
@indrora Any updates?

indrora (approved):

PR #835, opened 2023-04-19
Adding Jest tests into a test stage

jaysonmc:
1. Adding npm install to README (missing step)
2. Introducing a test stage (single action) to run the existing Jest tests and generate a CodeBuild report


typescript/codepipeline-build-deploy/app/buildspec.yaml
```diff
@@ -22,6 +23,8 @@ phases:
       - sed -i "s|EXECUTION_ROLE_ARN|${EXECUTION_ROLE_ARN}|g" taskdef.json
       - sed -i "s|TASK_DEFINITION_ARN|${TASK_DEFINITION_ARN}|g" appspec.yaml
       - cat appspec.yaml && cat taskdef.json
+      - mv appspec.yaml ../
```
indrora, line 26: Should be copy, not move, or the file should be there to begin with. 

I'd suggest outputting to a new file rather than modifying the file in-situ with sed.

jaysonmc: The sed part is not part of this PR. Though if we'd like to address, would make for a good new issue.

Regarding the move. I tried to reference the file not from root and it didn't work. I can take another swing at that one, though.

typescript/codepipeline-build-deploy/lib/codepipeline-build-deploy-stack.ts
```diff
@@ -205,14 +226,26 @@ export class CodepipelineBuildDeployStack extends cdk.Stack {
       stageName: "Source",
       actions: [
         new pipelineactions.CodeCommitSourceAction({
-          actionName: "CodeCommit",
+          actionName: "AppCodeCommit",
           branch: "main",
-          output: sourceArtifact,
+          output: SourceArtifact,
```
kaizio, line 231: Looks like you meant this to be lowercase
```suggestion
          output: sourceArtifact,
```

kaizio (approved):
LGTM

PR #842, opened 2023-04-29
Cloud trail log streaming opensearch serverless

rupeshtiwari:
Fixes # 

---


python/cloudtrail-streaming-to-opensearch-serverless/docs/streaming cloudtrail in OpenSearch.png
ericzbeard, line 1: Is that an approved image (the person at the laptop)?

python/cloudtrail-streaming-to-opensearch-serverless/docs/streaming cloudtrail in OpenSearch.xml
ericzbeard, line 1: What is this file?

python/cloudtrail-streaming-to-opensearch-serverless/stacks/opensearch_serverless_stack.py
```diff
@@ -0,0 +1,180 @@
...
+import boto3
+from aws_cdk.aws_s3_assets import Asset
+from constructs import Construct
+
+## Constants
+INDEX_NAME = "cwl"
+LOG_GROUP_NAME = "SvlCTCWL/svl_cloudtrail_logs"
+COLLECTION_NAME = "ctcollection"
+CWL_RETENTION = cwl.RetentionDays.THREE_DAYS
+ARN_IAM_USER = boto3.client("sts").get_caller_identity()["Arn"]
```
ericzbeard, line 27: Why is this necessary? If we make any network calls during synthesis, the app will not be deterministic, making it harder to test.

python/cloudtrail-streaming-to-opensearch-serverless/docs/streaming cloudtrail in OpenSearch.png
rupeshtiwari: That image of person on laptop is duty free image so we can use it. This is an architecture diagram.

python/cloudtrail-streaming-to-opensearch-serverless/docs/streaming cloudtrail in OpenSearch.xml
rupeshtiwari: This is the architecture diagram DRAWIO.XML file so that anyone can open the diagram on drawio and edit it.

python/cloudtrail-streaming-to-opensearch-serverless/stacks/opensearch_serverless_stack.py
rupeshtiwari: This is static API call that returns the ARN of the AWS user who is running the CDK. Instead of hard coding it we are getting this from the AWS config Credentials. This is deterministic call and always return idempotent value. I believe there is no harm using this line let me know what you think?

ericzbeard: Why do you need their role arn?

rupeshtiwari: Hi Eric, In Amazon OpenSearch we have to re-map the USER and Service Identity in order to provide a user access to browse the Amazon OpenSearch dashboard to visualize the data or search data. Therefore, I need the user ARN who is running cdk and I map that to the OpenSearch role so that after running CDK user can browse through the OpenSearch dashboard.

rupeshtiwari: In OpenSearch whoever user wants to login and access the dashboard that User ARN has to be mapped and I need that ARN. So normally this sample whoever will run will have some Admin user in mind who will after deploying open search will browse the dashboard and That user we have to map it with opensearch. In this sample CDK we are using the cdk deployment user's ARN. https://docs.aws.amazon.com/opensearch-service/latest/developerguide/fgac.html#fgac-mapping 

image (6)

Jon-AtAWS: There are 3 (service) policies in play - a network policy, a data access policy, and an encryption policy. These are related to, but distinct from IAM policies. You can find them all at the top of the file.

* The network policy specifies whether the API endpoint and the OpenSearch Dashboards endpoint are public, or accessed via a VPC endpoint 
* The encryption policy lets you specify AWS-managed, or user-managed KMS key for data encryption
* The data access policy specifies which user can access data in the collection.

We capture the user's identity (`ARN_IAM_USER = boto3.client("sts").get_caller_identity()["Arn"]`) and inline it into the data access policy. In the same policy, we enable Lambda to write to the collection.

```
  {{
    "Description": "Endpoint access for Lambda and for random querying",
    "Rules":[
        {{
          "ResourceType":"collection",
          "Resource":["collection/{COLLECTION_NAME}"],
          "Permission":["aoss:*"]
        }},
        {{
          "ResourceType":"index",
          "Resource":["index/{COLLECTION_NAME}/{INDEX_NAME}*"],
          "Permission":["aoss:*"]
        }}
    ],
    "Principal":["{ARN_IAM_USER}", "LAMBDAROLEARN"]
  }}
```

The policy must have the user's ARN or they won't have access to any data in the collection. As Rupesh said, we can make it a constant that the person deploying modifies before deploying. Grabbing it from boto is a convenience that works all the time when the person deploying the stack is the same as the person who needs to access the data.

What Rupesh detailed above about mapping for the backend is true for the provisioned version of the service, but doesn't apply to serverless.

For the record, these are the available Permissions (see: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-data-access.html)

## Collection permissions

* aoss:CreateCollectionItems
* aoss:DeleteCollectionItems
* aoss:UpdateCollectionItems
* aoss:DescribeCollectionItems
* aoss:*

## Index permissions
* aoss:ReadDocument
* aoss:WriteDocument
* aoss:CreateIndex
* aoss:DeleteIndex
* aoss:UpdateIndex
* aoss:DescribeIndex
* aoss:*

We could scope those down more so that Lambda has fewer permissions than the user by adding another clause to the data access policy. It seems like overkill to me, but you could definitely talk me into it.

rupeshtiwari: Thanks Jon I will connect with you.

rupeshtiwari: Hi Eric, when I removed the USER IAM Role ARN then I get unauthorized error since user will not have access to view the dashboard. 
image

Also I can ask user to enter their IAM role but personally I think that is not required for the sample application. However, if you need me to do that change I will be happy to make. I already added the note in README for the user to update that value if they wish todo.
image


Also Eric/Jon , I also find in other repositories we are retrieving `account_id` from the boto as well please check this repo. https://github.com/aws-samples/amazon-opensearch-service-monitor/blob/9c6a17c60bd746217205f1ffe6b2d452e40d2f65/opensearch/setupCWSubscriptionFilter.py#LL18C1-L18C66

ericzbeard (dismissed):

rupeshtiwari:
https://docs.aws.amazon.com/cdk/v2/guide/troubleshooting.html#troubleshooting_app_required

When synthesizing an AWS CDK stack, I get the message --app is required either in command-line, in cdk.json or in ~/.cdk.json

This message usually means that you aren't in the main directory of your AWS CDK project when you issue cdk synth. The file cdk.json in this directory, created by the cdk init command, contains the command line needed to run (and thereby synthesize) your AWS CDK app.

ericzbeard (dismissed):

ericzbeard (dismissed):

ericzbeard (approved):

PR #851, opened 2023-05-19
Added Datasync CDK example

kawsark:
Added a CDK example to support AWS DataSync S3 use-cases such as backups, bucket consolidation, and data lake creations.

Fixes [# 852](https://github.com/aws-samples/aws-cdk-examples/issues/852)

---


mmaeng:
In order to pass the build tests it looks like you will have to provide at least placeholder `S3_datasync_locations` contexts so the synth will pass.

kaizio:
Stale.

PR #861, opened 2023-06-14
feat(SFTP Server): AWS Transfer Family: CDK example to create a fully managed SFTP server to transfer files directly into and out of Amazon S3

### LINKED ISSUE #553: AWS Transfer Family: CDK example to create a fully managed SFTP server to transfer files directly into and out of Amazon S3
### General Information

The AWS Transfer Family offers fully managed support for the transfer of files over SFTP, FTPS, and FTP directly into and out of Amazon S3 or Amazon EFS. Let's create a CDK example of how to configure an SFTP server using AWS Transfer Family

### Proposed Solution

The CDK solution will spin up the following components:

- Transfer Family SFTP server
- S3 bucket
- IAM role and policy
- Username and Password stored in Secrets Manager

### Language

Typescript, Python, .NET

### Other information

_No response_

### Acknowledge

markusl:
CDK example to create a fully managed SFTP server to transfer files directly into and out of Amazon S3

Fixes #553

---


typescript/aws-transfer-sftp-server/aws-transfer-sftp-server.ts
```diff
@@ -0,0 +1,152 @@
...
+      assumedBy: new iam.ServicePrincipal('transfer.amazonaws.com'),
+      description: 'IAM role used by AWS Transfer for logging',
+      inlinePolicies: {
+        loggingRole: new iam.PolicyDocument({
+          statements: [new iam.PolicyStatement({
+            actions: [
+              'logs:CreateLogGroup',
+              'logs:CreateLogStream',
+              'logs:DescribeLogStreams',
+              'logs:PutLogEvents',
```
indrora, line 63: My critique here would be that I would rather see PutLogEvents scoped down to a single LogGroup.

typescript/aws-transfer-sftp-server/aws-transfer-sftp-server.ts
```diff
@@ -0,0 +1,152 @@
...
+      maxAzs: 2,
+      natGateways: 0,
+    });
+
+    this.incomingDataBucket = new s3.Bucket(this, 'IncomingDataBucket', {
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      bucketName: `sftp-server-data-bucket-${this.account}-${this.region}`,
+      encryption: s3.BucketEncryption.KMS_MANAGED,
+      enforceSSL: true,
+      // Do not use for production
```
indrora, line 48: Don't put it in an example. 

This bucket should be passed up through the `SftpServerStackProps`.

typescript/aws-transfer-sftp-server/main.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// Follow the setup process at https://docs.aws.amazon.com/cdk/v2/guide/environments.html
+const props: cdk.StackProps = {
+  env: {
+    account: process.env.CDK_DEFAULT_ACCOUNT,
+    region: process.env.CDK_DEFAULT_REGION,
+  }
+};
+
+const userPublicKeys = [
```
indrora, line 13: Nit: Something like this should be from Parameter Store or similar, and should be documented. For this example, hardcoding the SSH key is fine but customers should use context values, parameter store, secrets manager, etc. for holding things like this.

indrora (dismissed):
Overall this looks good but has a few nitpicks here and there. 

In the long run, this would make a better L2 construct (or L3, if the maturity of some components can be worked out) and would be good to submit as such, rather than as an example given the amount of specific work that has to be done. 

I would appreciate a few more comments -- things are a little sparse -- explaining things like the IAM roles and users that are added. 

w/r/t SSH keys: Such things for development are fine hardcoded but should be, in the long run, attached to something like Parameter Store or Secrets Manager, depending on the secret-ness of them. Parameter store is optimal for things like public keys but that's beside the point here. 

With that said, I want to accept this PR but need to see a few more comments and the S3 bucket being "outside" of the stack as it were.

typescript/aws-transfer-sftp-server/aws-transfer-sftp-server.ts
markusl: Is this possible with IaC? CfnServer depends on the role which needs to be defined before we know the specific log group name (which is named after the CfnServer).

indrora (dismissed):

indrora (approved):

PR #874, opened 2023-06-30
feat(build): Parallelize the build step

indrora:
This uses the github matrix to run each language build in parallel.

it does so by filtering out the files that are only used by that language. 

In addition to parallelizing the builds, it gives each Python build its own venv, as Python is the only language that does not have local packages by default. 

---


kellertk (approved):
TODO: Investigate the use of caching in the action/setup* packages. This may improve subsequent execution times further.

PR #894, opened 2023-07-25
Fix(gh actions): Add back DO_NOT_TEST check

kaizio:
Looks like this was removed at some point.


pahud:
I am working on the build-typescript.sh at https://github.com/aws-samples/aws-cdk-examples/pull/893

Will include this logic into my PR.

pahud:
https://github.com/aws-samples/aws-cdk-examples/blob/6b1bc9a707af4c748bbe5341e852fb3fe58ff98b/scripts/build-typescript.sh#L17

I think we can close this PR. What do you think? @kaizio

PR #943, opened 2023-10-12
Fix(ts): Fix aws-cdk-lib and deprecated method/properties

kaizio:
After updating to latest aws-cdk-lib theses are the fixes for simple deprecated method/property fixes.

---


bilalmughal:
If we are updating the `aws-cdk-lib`, why don't we bump it to the latest available version, `2.100.0`, instead of `2.99.1`?"

kaizio:
CDK has been releasing new version, at least in the past month, every 4-5 days. I started this about a week back. I am planning on following this up with workflow that will run once a month that will automate the version upgrades to match current cdk init. It will also report any deprecated methods so that we can go back and fix them.

Issue #765, opened 2022-10-26
New Pattern/ Example Idea: EC2 with SSH Access to EMR Cluster


tommywhelan:
### Describe the feature

I am thinking of making a pattern that spins up a basic EMR cluster, with an accompanying EC2 Instance that has SSH access to the cluster, for secure access.

### Use Case

Frequently, I will spin up an EMR cluster to test something out, but it often takes a while to actually set up the SSH access so that I can appropriately use the cluster. 

With this pattern, EMR developers will quickly be able to spin up their EMR cluster and have access to it in order to test out functionality and develop on EMR.

### Proposed Solution

I made this rough diagram/ sketch to show what the eventual implementation might look like...

![emr_cdk](https://user-images.githubusercontent.com/97986883/198090761-356c49b6-fbbc-4e5b-944b-80e29f57cfd7.png)

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Language

Python

tommywhelan:
Just following up!

Does this seem to be a valid pattern that would be of use for the repo?

tommywhelan:
I have already developed the code for this pattern, and could easily it include it in this repo (with some checks of course)...

Is there anyone I can speak to/ communicate this with??

kaizio:
Yes this would be a good example for this repo.

tommywhelan:
@kaizio What would be the next steps? Should I make a PR my new pattern?

PR #954, opened 2023-10-27
Add New Python Pattern: SSH-able EMR cluster via EC2 Instance

tommywhelan:
PR for validated idea from existing issue that I created:
https://github.com/aws-samples/aws-cdk-examples/issues/765

The pattern creates an EMR cluster that is in a Private Subnet, which can be SSH-ed into in a secure fashion via a EC2 instance in the VPC's public subnet.

This pattern provides an accelerator for developing on EMR, by reducing the time/ # of steps in which a user can gain secure access to the cluster.

---


python/ssh-into-emr-cluster/emr_pattern/emr_pattern_stack.py
```diff
@@ -0,0 +1,194 @@
...
+            ),
+            security_group=ec2_security_group,
+            key_name="emr_test"
+        )
+
+        # S3 Bucket used for: logging, storing emr scripts, Spark job input data, and Spark job output data
+        s3_bucket = s3.Bucket(
+            self,
+            "s3_bucket",
+            bucket_name=f"emr-cdk-pattern-{Aws.ACCOUNT_ID}-{str(uuid4())[:5]}",
```
kaizio, line 100: This is going to cause a bucket to get created on each synth/deploy even you have no changes.

python/ssh-into-emr-cluster/requirements.txt
```diff
@@ -0,0 +1,2 @@
+aws-cdk-lib==2.49.0
```
kaizio, line 1: update to latest

python/ssh-into-emr-cluster/emr_pattern/emr_pattern_stack.py
tommywhelan: @kaizio Thanks for checking through code + finding this bug, good callout! I have made another commit to my fork removing this UUID section for bucket name, as it shouldnt be necessary.

python/ssh-into-emr-cluster/requirements.txt
kaizio: Would you be able to update this to latest. 2.49 is pretty old released back in Oct 27, 2022.

tommywhelan: ^sure! this pattern was worked out a while back, so I guess that's why it's outdated

Let me upgrade CDK version, retest, and follow up here today

tommywhelan:
@kaizio Fixed! 

Just tested with new CDK version and worked smoothly. From my end, this pattern is good to go

kaizio (commented):
:+1:

PR #963, opened 2023-11-19
feat(ts): Aspects example

### LINKED ISSUE #962: Add CDK Aspect Implementation Example
### Describe the feature

Suggesting a CDK Aspect implementation example to be added as many users might not have experience using the feature.

### Use Case

I have used Aspects for reporting, security scanning, and applying bulk customizations to constructs (Though I would only recommend the latter as a last resort). 

### Proposed Solution

I have created a sample project implementing an aspect that does the following:
- Set a default ReservedConcurrentExecutions property in lambda functions across a scope if the property is null/undefined.
- Logs construct paths and actions taken to give users an idea of the aspect's behavior.

### Other Information

While the contributing guidelines discourage escape hatches, I still believe this example is relevant to show users how to implement a custom aspect.

### Acknowledgements

### Language

Typescript

fosterdrummer:
Fixes #962

Adding an example implementation and application of an Aspect.

The sample project implements an aspect that does the following:
- Sets a default ReservedConcurrentExecutions property in lambda functions across a scope if the property is null/undefined.
- Logs construct paths and actions taken to give users an idea of the aspect's behavior.

Code comments and documentation are added to provide context around how aspects work and steps to implement a custom aspect.

---


kaizio (approved):
Thanks look good.

PR #978, opened 2023-12-19
Provision OpenSearch Cluster in private subnet for private setup and access via NGINX Proxy

ki2a:
Currently with given solution, OpenSearch cluster is getting provisioning in public subnet which makes complete cluster public and accessible even without NGINX proxy instance. For VPC based OpenSearch cluster, it should be provisioned in private cluster with security group allow port 80 & 443 inbound connection from NGINX EC2 security group. 

Fixes # 

---


kaizio (approved):
LGTM

PR #989, opened 2024-01-30
initial commit for eks private cluster

madrian:
Fixes # 

---

This is an example to create an EKS cluster in CDK with in isolated subnets with no internet or nat gateways.


java/eks/private-cluster/src/main/java/com/amazonaws/cdk/examples/EksPrivateClusterStack.java
```diff
@@ -0,0 +1,228 @@
...
+
+    for (InterfaceVpcEndpointAwsService e : endpoints) {
+      vpc.addInterfaceEndpoint(
+          e.getShortName(),
+          InterfaceVpcEndpointOptions.builder().service(e).subnets(subnets).build());
+    }
+
+    vpc.addGatewayEndpoint(
+        "s3", GatewayVpcEndpointOptions.builder().service(GatewayVpcEndpointAwsService.S3).build());
+  }
```
paulosergio-jnr, line 219: WDYT of advising the developer (through a documentation or comment) that this will considerably increase the costs of networking and suggest adding only the endpoints that make sense for their infrastructure?

madrian: Thanks for the feedback. I have updated the README with a note on the VPC Endpoint pricing. After further testing, have removed CloudFormation and EventBridge in the endpoints as they are not needed.

kaizio (approved):

PR #1002, opened 2024-03-12
Adding new sample: api-gateway-lambda-response-static-page

fanhongy:
build a new sample that use API Gateway, Lambda function and DynamoDB. The Lambda function will query a preloaded DDB and send back the result as a html page in response as a static page. 

---


kaizio:
@fanhongy This looks very similar to an example that we already have in python [apigw-http-api-lambda-dynamodb-python-cdk](https://github.com/aws-samples/aws-cdk-examples/tree/master/python/apigw-http-api-lambda-dynamodb-python-cdk)

I do see that you are returning a html in your handler but that is more of an AWS API Gateway response difference than demonstrating CDK concepts.

Please reopen/message me if you think otherwise.

PR #999, opened 2024-03-04
chore(typescript): update sample to use nodejs 18

crhorne-aws:
Fixes #961 

- Updates lambda functions' nodejs runtime for `typescript/api-gateway-parallel-step-functions` sample.
- Updates construct props to account for deprecation warnings

---

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license].

[Apache 2.0 license]: https://www.apache.org/licenses/LICENSE-2.0

PR #1007, opened 2024-03-14
chore(typescript/lambda-layer-stack.ts): update sample to use node 18 (#999)

allipry:
Fixes # 

---


kaizio:
@allipry  So the big thing with going to node18_X is that we swap from AWS SDK v2 to v3. The code for lambda also needs to be updated if the code imports the AWS SDK.
https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/migrating.html

PR #1011, opened 2024-03-22
Add example of @aws-cdk/route53-resolver-alpha and CfnResolverEndpoint

cam8001:
A mixed example making use of both an alpha L2 module, `@aws-cdk/route53-resolver-alpha`, which implements the DNS firewall feature of Route 53 Resolver, and `aws-cdk-lib/route53-resolver.CfnResolverEndpoint`.

Typescript only.

This:
- sets up a VPC
- deploys the DNS firewall with some basic rules to block `example.com` and `example.net`
- Creates an inbound resolver to serve requests from inside the VPC
- Creates an outbound resolver to serve requests from outside the VPC.

### DIFF
--- a/typescript/r53-resolver/lib/vpc.ts
+++ b/typescript/r53-resolver/lib/vpc.ts
@@ -1,0 +1,22 @@
+import * as cdk from "aws-cdk-lib";
+import { Construct } from "constructs";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+
+export class R53ResolverVPC extends Construct {
+  public vpc: ec2.Vpc;
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+    // @see https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Vpc.html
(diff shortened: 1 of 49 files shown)


kaiz-io (approved):
:+1:

typescript/r53-resolver/lib/vpc.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Construct } from "constructs";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+
+export class R53ResolverVPC extends Construct {
+  public vpc: ec2.Vpc;
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+    // @see https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Vpc.html
+    this.vpc = new ec2.Vpc(this, "R53ResolverTestVPC", {
+      ipAddresses: ec2.IpAddresses.cidr("10.24.34.0/23"),
```
cam8001, line 11: can we make a parameter or envvar?

PR #1015, opened 2024-03-24
Cdk pipeline cross account example

rushali-aws:
Adding CDK example to create a cross account CodePipeline with two stages , Source stage as CodeCommit repository from another account and CodeBuild stage is same as that of Pipeline account. 

We have few examples with Cross account configuration however we do not have any existing example with cross account source as a source.


kaiz-io:
Build is failing to compile. Make sure you are able to run.

`npm run --if-present test`
`npx cdk synth`

PR #1034, opened 2024-05-13
Updated READMe with information about avoiding potential 403 status code

hellofadude:
Fixes # 1029 

Updated READMe for static-site because I was getting 403 status code, because I was using an administrative account that was not the root account. I figured rather than put in a fix directly in the code, as not everyone will be in the same situation, an update to the READMe.md would be a better solution.
---


kaiz-io:
Have you tried using a user that had AdminAccess and not SCP/Permissions Boundary? You should not need to use the root user account, as that is not a best practice for AWS governance.

Issue #961, opened 2023-11-08
Update Lambda examples to use NODEJS_18_X


trivikr:
### Describe the feature

Update Lambda examples to use NODEJS_18_X

### Use Case

There are multiple examples on Lambda Node.js 10.x and 12.x which are no longer supported by Lambda:
* 10.x https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_10_X&type=code
* 12.x https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_12_X&type=code

And per [official documentation](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html), Node.js 14.x and 16.x will be following deprecation phase 1 on Nov 27, 2023 and Jun 12, 2024 respectively. These are also used in CDK examples:
* 14.x https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_14_X&type=code
* 16.x https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_16_X&type=code

### Proposed Solution

Upgrade all examples to use NODEJS_18_X

Announcement blog post https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Language

Typescript

trivikr:
Closing this issue as there are no lambdas with Node.js 10/12/14/16
* https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_10_X&type=code
* https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_12_X&type=code
* https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_14_X&type=code
* https://github.com/search?q=repo%3Aaws-samples%2Faws-cdk-examples%20NODEJS_16_X&type=code

PR #1046, opened 2024-05-30
chore: update Lambda Node.js to 20.x

trivikr:
Noticed in: https://github.com/aws-samples/aws-cdk-examples/issues/961

Node.js 20.x is the latest version of Lambda https://aws.amazon.com/blogs/compute/node-js-20-x-runtime-now-available-in-aws-lambda/

---


kaiz-io:
The tests failed make sure that you can run `yarn run test` or ` npm run --if-present test` and that cdk synth works.

kaiz-io:
So just fixed an issue that might have been affecting one error that you were getting. Currently the build is failing because in the project `typescript/api-gateway-lambda-token-authorizer/` the test that is configured is failing because the snapshot is different.

Here is the command to run in the director to replicate the issue. `npm run --if-present test`

PR #1047, opened 2024-06-14
Added new example - Serverless Multi-region tokenizer with expiry

simon007-aws:
Fixes # 

---


simon007-aws:
I am unsure why this is failing. I suspect it is because this is a multi-region stack, and the expected constructs are not in place.

simon007-aws:
Initial commit to aws-samples

kaiz-io:
Great work. This would be better as a repo on its own with the complexity of what it is demoing. As it already does at https://github.com/simon007-aws/serverless-multi-region-tokenizer

Examples in this repo should be mainly showing CDK examples that are simple and straight forward. Showing how to implement various services and maybe some pipeline integration with others. This is more of a solution/methodology to using tokenization in multi region with serverless..

PR #1072, opened 2024-09-13
Add Amazon Connect CDK example in Python

kiryanko-aws:
This PR adds an example in Python that deploy an Amazon Connect instance and configures it for use, including setting up a call recordings S3 Bucket, a scheduled reports S3 Bucket, and a Firehose Delivery Stream for CTR. Moreover, the solution assigns a phone number to the instance and sets up the hours of operation.

---


python/amazon-connect/requirements.txt
kaiz-io, line 1: You are missing cdk_nag which is causing the build to fail.

kaiz-io (commented):
Fix the build with including cdk_nag and everything else looks good. Thanks!

python/amazon-connect/requirements.txt
kiryanko-aws: Apologies, that has been added now.

kaiz-io (approved):
Looks good

PR #1087, opened 2024-10-17
add stepfunctions with external asl definition file

pedramha:
This PR adds an example of implementing stepfunctions with external asl(Amazon States Language) file. This can be useful in usecases where it is desired to leverage local development flow of the stepfunction statemachines,e.g. AWS toolkit for VScode, and directly import the asl file into your project without rewriting it in CDK.

---


typescript/stepfunction-external-definition/lib/stepfunction-external-definition-stack.ts
```diff
@@ -0,0 +1,51 @@
...
+    const stateMachine= new stepfunctions.StateMachine(this, 'MyStateMachine', {
+      stateMachineType: stepfunctions.StateMachineType.EXPRESS,
+      definitionBody: stepfunctions.DefinitionBody.fromString(workflow.toString()),
+      logs: {
+        destination: logGroup,
+        level: stepfunctions.LogLevel.ALL,
+        includeExecutionData: true,
+      },
+    });
+    stateMachine.addToRolePolicy(new iam.PolicyStatement({
```
kaiz-io, line 28: This appears to want to give access for the state machine to be able to write to the log group. Instead of creating a permission policy with a * in and having to track that you have all the correct API's, please change to using [LogGroup.granteWrite](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.LogGroup.html#grantwbrwritegrantee).

It would look something like `logGroup.grantWrite(stateMachine.role);`

kaiz-io (changes requested):
Please correct the policy correction. Otherwise it looks good. Thanks for the contribution.

pedramha:
Adjusted the permission setting accordingly, thanks for the input @kaiz-io.

kaiz-io (approved):
:rocket:

PR #1106, opened 2024-12-06
add readme, adjust cdk.json, add platform to Dockerfile, use one NAT

klang:
Readme file was missing, it's added with explanation of functionality.

NATGateways set to 1 to reduce cost and deployment time for the example.

Platform specified in Dockerfile to accommodate deployments from varied platforms. Fargate will default to `linux/amd64`. Personal desktop machines may default to some `arm` variant.

`cdk.json` was missing the correct specification to build the app.

---


kaiz-io (approved):
Thanks

PR #1153, opened 2025-04-04
feat(Issue#791) - CDK Sample: Event Bridge mesh with CDK

### LINKED ISSUE #791: CDK Sample: Event Bridge mesh with CDK
### Describe the feature

A CDK way to set up a Event Bridge Mesh, where you relay the messages from one Event Bridge in a producer account to another Event Bridge in a consumer account

### Use Case

(Based on Enterprise Customer necessity). Customers that have implemented IaC usually have different teams controlling Producer and Consumer accounts.
They want the consumers to be able to control which messages they want to process and where they go, and evolve that logic over time, without having to ask the Producers to modify the event rules each time they want to go to production with a change. 
For this it's better to separate Event Bridges in producers and consumers. Producers will have to do a one-time setup to forward events to the Consumer, and then consumers can evolve the way they filter and route those messages in their applications.

### Proposed Solution

The sample would be this
EventBridge

The Reference architecture would be this
EventBridge(1)

### Other Information

If you use a single EventBridge for Producers and Consumers, that are managed by different teams, you lose the agility and independence for each team, as the Consumer teams will be constantly requiring modifications on the rules/targets to the producer team. 

### Acknowledgements

### Language

Python

jfan9:
Fixes #791   
1. Created single consumer and multiple consumers scenarios in Typescript
2. Created single consumer and multiple consumers scenarios in Python
3. Tested both solutions which worked fine
4. Create a README.md for the architecture design and instructions 
---


jfan9:
This PR used typescript, python's PR will be submitted later.

jfan9:
Python example has been added

jfan9:
@kaiz-io, both typescript and python examples are added for issue #791, please kindly review it, thank you

kaiz-io (approved):
:rocket:

